PR #6, opened 2015-09-17
Update README

agirish:


Agirish:
(3) is already mentioned. Java and Maven version information is tied to Apache Drill pre-requisites.

Lets add a new INSTALL.md file and move much of these from README to it.

PR #9, opened 2015-10-01
Metadata Caching

rchallapalli:
Vicky,

This pull request contains 
1. Metadata Caching Tests
2. Fix for  "DATA_READ ERROR: Selected column 'dir0' must have name 'columns' or must be plain '*'"

Below is the functional run with one failure which I think is not new
http://10.10.104.91:8080/job/Framework%20Regression/49/consoleFull

Let me know your comments
- Rahul

### DIFF
--- a/framework/resources/Datasources/tpcds/createTablesJson.sql
+++ b/framework/resources/Datasources/tpcds/createTablesJson.sql
@@ -1,6 +1,6 @@
-/* Author: Abhishek Girish */
-
 alter session set `store.format` = 'json';
+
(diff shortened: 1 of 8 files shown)


vmarkman:
I verified that precommit run was successful (one known failure: https://issues.apache.org/jira/browse/DRILL-3480), tests were executed against latest commitID.

Ship it !

Agirish:
Hey Rahul, since the run links aren't publicly accessible, we thought it would make sense to avoid doing that, for now. We should decide on a better way to share run logs soon.

PR #18, opened 2015-10-08
Bug verifcation tests

vmarkman:
@Agirish - can you please review ?

Thank you,
Vicky.


Agirish:
Thanks for transplanting these Vicky! Since there are so many commits, give me some time to take a look. However, since these have already been previously reviewed, I don't expect any issues.

vmarkman:
yes, these tests are replanted "as is" from the old branch. Thank you !

Agirish:
Looks good. +1

PR #23, opened 2015-10-12
Fix expected plan verification failures

rchallapalli:
Abhishek,

I fixed the expected plan verification failures. Can you please review this pull request?
- Rahul


Agirish:
So this is the one where you basically had to fix it twice to essentially revert it back to original right? Looks good. +1

rchallapalli:
Yup!

Thanks Abhishek

PR #26, opened 2015-10-14
Onelevel

rchallapalli:
Team,

I got rid of the 2 level maven structure in the test framework. This has been a source of confusion for sometime now. Let me know if you all feel comfortable with this change. I got 2 successful runs one on each cluster which runs the new test framework.

http://10.10.104.91:8080/job/Framework%20Regression/99/console
http://10.10.104.91:8080/job/Framework%20Regression/100/console

Rahul


Agirish:
This is a welcome change! We should try and get this in soon. Since it will have impact on existing setups, lets discuss offline.

PR #28, opened 2015-10-21
Drill 3947

rchallapalli:
Abhishek,

I verified drill-3947 and added some test cases. The dataset for these test cases should contain more than 4096 auto partitioned files to reproduce the issue. So there are lot of new small data files added as part of this commit. The functional run passed without any issues. Let me know your thoughts
- Rahul


Agirish:
Hey Rahul, I think we should take a look at this together in person. Github doesn't show me any changes made, since the number of files are so many. I tried looking at your branch, but its just the same.

rchallapalli:
Sure!

Agirish:
As discussed, looks good. +1

PR #31, opened 2015-10-26
Metadata Caching + Partition Pruning tests

rchallapalli:
Abhishek,

Added more metadata caching tests. Got a clean functional run. Can you take a look?
- Rahul


rchallapalli:
The first commit is all data files necessary and the second commit contains the actual tests

Agirish:
Hey Rahul, in the future I thought we could include "drop table" statements in the scripts. (example: framework/resources/Datasources/metadata_caching/ctas_auto.ddl). For now, since you are using a script for this, we must be fine. 

Overall looks good. +1

rchallapalli:
Thanks Abhishek! You are right. Lets start using the drop table from now on.

PR #33, opened 2015-10-28
Hive : added integer and text partition tests for parquet and text data

rchallapalli:
Vicky,

I added data verification tests around hive dynamic partitions. Added tests for int, string & date partition types for text and parquet serde's. Will add the plan verification tests later after we finalized on the approach. Let me know your thoughts.
- Rahul


vmarkman:
Rahul,

Good tests, ship it. Please research NULL issue as we discussed (TPCH schema does not have NULLs)

Vicky.

rchallapalli:
Thanks for your review vicky. I will add tests around NULL's in dynamic partitions.

Rahul

PR #39, opened 2015-11-06
DRILL-4205: Change plan to indicate use of metadata cache file

mehant:
@rchallapalli please review.


rchallapalli:
Mehant,

The expected value's regex might not match what's returned from drill

-1

- Rahul

PR #40, opened 2015-11-07
Updating startup scripts

agirish:
Hey @cchang738, 

As discussed, I've updated the startup scripts to be more smart. Can you take a look? I have a regression run going on - will let you know if they pass. 

Thanks,
Abhishek


Agirish:
Added the exit statement.

Agirish:
Got a successful run. Discussed with Chun. Merging.

PR #43, opened 2015-11-10
Failexit

rchallapalli:
Chun,

I made changes to the test framework so that we fail immediately after an Execution Error happens. This would help us preserve the state of the drill cluster when the error happened. Let me know your thoughts
- Rahul

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -145,6 +145,9 @@ public class TestDriver {
     @Parameter(names = {"-h", "--help"}, description = "show usage", help=true)
     public boolean help = false;
 
+    @Parameter(names = {"-e"}, description = "Exit on Failure", required=false)
+    public boolean exitOnFailure = false;
+
     @Parameter(names = {"-x", "--exclude"}, description = "Dependencies to exclude", required=false)
     public String excludeDependencies = null;
 
--- a/framework/src/main/java/org/apache/drill/test/framework/CancelingExecutor.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/CancelingExecutor.java
@@ -70,7 +70,7 @@ public class CancelingExecutor implements AutoCloseable {
             try {
               future.get(timeout, TimeUnit.SECONDS);
               if (task instanceof DrillTestJdbc) {
-                if (((DrillTestJdbc) task).getTestStatus() == TestVerifier.TestStatus.EXECUTION_FAILURE) {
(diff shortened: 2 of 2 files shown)


rchallapalli:
This changes undoes one of sudheesh's change. It adds back logging to console when datagen scripts are executing. Without this message it is very hard to debug when some data gen script fails.

cchang738:
LGTM. But like to see a test run to confirm this actually works. If you have tested locally and verified its effectiveness, go ahead and merge. Thanks.

sudheeshkatkam:
This does not capture the jstack?

sudheeshkatkam:
Sorry, but -1. An executor that shuts down the pool that it is created from?

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -402,7 +405,7 @@ public void run() {
   private static void runGenerateScript(DataSource datasource) {
     int exitCode = 0;
     String command = CWD + "/resources/" + datasource.src;
-    LOG.debug("Running command " + command);
+    LOG.info("Running command " + command);
```
sudheeshkatkam: Why isn't the "LOG.error(...)" below sufficient?

rchallapalli: Since we are running the datagen scripts in parallel, we want to know what else was running when this failure happened. On the other hand console output from different threads could lead to some confusion.

I am inclined to have more information displayed.

sudheeshkatkam: Sounds good.

Agirish: Just wanted to add that I also think we should display what scripts are running. Having more information would help. If this needs to be part of debug, then our Jenkins scripts need to be updated to run in debug mode by default.

vmarkman: +1 on adding more information. It's really hard to debug when something goes wrong.

sudheeshkatkam: which is what debug level is meant for ;)

rchallapalli: I will isolate the change to add more debugging and create another pull request. 

The original issue addressed in this is a hacky one :) as sudeesh pointed out and the usefulness is also questionable as the state of the drillbit at the time of failure is not accurately captured (drill cancels running queries if client terminates connection)

PR #45, opened 2015-11-13
limit 0 tests with TPCH, taking out failing tests

vmarkman:


Agirish:
So in summary 
- 1 new test suite with 15 tests have been added (TPC-H queries with limit 0). Verified the test ID and tests
- 1 test previously failing with verification issue (due to Decimal casts) was fixed. 
- Some tests were renamed - nothing much to verify here. 

Looks good. +1

vmarkman:
Abhishek,

Really appreciate you spelling out what is done, I should do it next time.

Thank you for the review,
Vicky.

PR #47, opened 2015-11-13
Hive Dynamic Partition Pruning tests

rchallapalli:
Abhishek/Vicky,

Added more hive partition pruning tests. Can you take a look?
- Rahul


vmarkman:
Awesome tests. If you are testing nulls, always try coalesce and nvl. Guaranteed to break :)
+1 if you ran regression and there were no failures caused by your tests.

rchallapalli:
Thanks vicky! I got a functional run with expected failures. I will also add more tests with coalesce and nvl functions.

PR #50, opened 2015-11-17
dfs impersonation tests

cchang738:
All impersonation tests are defined as the category "impersonation" in each test definition json file. If you do not specify "impersonation" as a group you want to run, impersonation tests will not be picked up under Functional.

Impersonation tests do require two regular users (user1, user2) and a user group (impersonation) already exist on the cluster running the test.


Agirish:
Have been having detailed discussions with Chun on this from a few days. Code change + tests looks good. +1

cchang738:
Thanks Abhishek for the review.

PR #51, opened 2015-11-18
change verification to regex for impersonation groupusershowfiles test

cchang738:
Verification need to be changed to regex and expected result also need to be modified for one impersonation test case.

Please review.


Agirish:
Looks good. +1

cchang738:
Thanks Abhishek.

PR #52, opened 2015-11-19
Tests to cover querying auto partitioned parquet data that is stored in nested directory that has a long name.

khurram-faraaz:
Vicky, can you please review these new tests. These are cover querying auto partitioned parquet data that is stored in nested directory that has a long name.

PS: Don't go by the commit message, "Add redundant tests", it is a typo, it should have been "Remove redundant tests".

Thanks,
Khurram


vmarkman:
Ship it. Nice. Verified that precommit run was clean and was executed with the latest commit id.

kfaraaz:
Thanks!

PR #54, opened 2015-11-20
MD-569: Framework now able to run a set of sql statement before test starts

cchang738:
This is a nice enhancement to framework to provide the functionality documented in MD-569. Lay groundwork to handle MD-158.

Pre-commit test run can be viewed here. Please review.

http://10.10.104.91:8080/job/Impersonation%20Regression/7/console

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -139,7 +139,7 @@ public class TestDriver {
     public int iterations = 1;
     
     @Parameter(names = {"-f"}, description = "filename", required=false)
-    public String beforeTestQueryFilename = "before-test.sql";
+    public String beforeRunQueryFilename = "before-run.sql";
     
     @Parameter(names = {"-d"}, description = "generate data", required=false)
     public boolean generate = false;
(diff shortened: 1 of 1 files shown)


Agirish:
Hey Chun, 

This is a good feature addition - should also help when we plan on extending this to per suite level. 

Can we move the "before-test.sql" file inside the `scripts/sqlScripts` directory? And possible rename it to something different - say like "initialize.sql" - since this isn't test specific, but for the whole run. 

Also, lets avoid on sharing private links - i had a look at the test run results - they look good.

cchang738:
Thanks for the review and feedback. I feel the location of the file should not be in the scripts directory. It's not a script. Also, it's more logical to have the file in the same directory where the run.sh is (where you start the run). I do agree on the suggestion on changing the file name. How about "before-run.sql"? I will add corresponding "after-run.sql", "before-test.sql" and "after-test.sql" in a future enhancement.

cchang738:
Changed default pre-run query filename to "before-run.sql". Looking for a +1. :)

Agirish:
+1. I would still prefer if the query script file was moved to the sqlScripts directory, to keep the framework dir clean. But I guess we could do it another time.

cchang738:
Ahbishek, I will sync up with you on your point offline. Thanks for review.

PR #57, opened 2015-11-24
Test Framework talks with zookeeper

rchallapalli:
Chun,

I made the change so that the test framework talks with zookeeper instead of a single drillbit. I got a clean functional run today and hanifi got 4 consecutive functional runs in the past. Let me know your thoughts
- Rahul


rchallapalli:
Chun,

I will add the new parameter to the drillTestProperties template file. 

Regarding backward compatibility, I think the default should always be "connecting via zookeeper". I will not commit this change until tomorrow. So that gives time for others to comment. What do you think?

Rahul

PR #59, opened 2015-11-25
file before-test.sql changed to before-run.sql

cchang738:
Rahul, please review and merge if no issue. Thanks.


cchang738:
pre-commit test passed. +1

PR #62, opened 2015-12-02
MD-575 support "limit 0" type of tests

cchang738:
Please refer to MD-575 for feature and implementation summary. Pre-commit run at the following link.

http://10.10.104.91:8080/job/Framework%20Regression/234/consoleFull

Please review. Thanks.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
@@ -52,6 +52,7 @@ public class DrillTestJdbc implements DrillTest {
   private TestMatrix matrix;
   private Thread thread;
   private List<Integer> types;
+  private List<Integer> nullability;
   private List columnLabels = new ArrayList<String>();
   private Random rand = new Random();
 
@@ -90,12 +91,17 @@ public class DrillTestJdbc implements DrillTest {
       
       query = queries[mainQueryIndex];
       executeQuery(query);
-        
+      
       testVerifier = new TestVerifier(types, query, columnLabels, matrix.verificationTypes);
       if (query.startsWith("explain") || matrix.verificationTypes.get(0).equalsIgnoreCase("regex")) {
         setTestStatus(testVerifier.verifyTextPlan(matrix.expectedFile, outputFilename));
       } else {
         setTestStatus(testVerifier.verifyResultSet(matrix.expectedFile, outputFilename));
+      }
+      
+      if (modeler.type.equalsIgnoreCase("limit 0")) {
+    	  String limitZeroQuery = "select * from (" + query + ") t limit 0";
+    	  executeLimitZeroQuery(limitZeroQuery);
       }
     } catch (VerificationException e) {
       fail(TestStatus.VERIFICATION_FAILURE, e);
@@ -149,7 +155,6 @@ public class DrillTestJdbc implements DrillTest {
   }
 
   private void executeQuery(String query) throws IOException, SQLException {
-
     outputFilename = Utils.generateOutputFileName(matrix.inputFile, modeler.testId, false);
     BufferedWriter writer = new BufferedWriter(new FileWriter(new File(
             outputFilename)));
@@ -191,13 +196,13 @@ public class DrillTestJdbc implements DrillTest {
     }
     
     try {
+      types = Lists.newArrayList();
+      nullability = Lists.newArrayList();
       int columnCount = resultSet.getMetaData().getColumnCount();
       for (int i = 1; i <= columnCount; i++) {
         columnLabels.add(resultSet.getMetaData().getColumnLabel(i));
-      }
-      types = Lists.newArrayList();
-      for (int i = 1; i <= columnCount; i++) {
         types.add(resultSet.getMetaData().getColumnType(i));
+        nullability.add(resultSet.getMetaData().isNullable(i));
       }
 
       LOG.debug("Result set data types:");
@@ -241,6 +246,54 @@ public class DrillTestJdbc implements DrillTest {
     }
   }
 
+  private void executeLimitZeroQuery(String query) throws IOException, SQLException {
+	if (getTestStatus() == TestStatus.CANCELED || getTestStatus() == TestStatus.EXECUTION_FAILURE) {
+		return;
+	}
+	
+	BufferedWriter writer = new BufferedWriter(new FileWriter(new File(outputFilename),true));
+	
+    Statement statement = connection.createStatement();
+    ResultSet resultSet = statement.executeQuery(query);
+    
+    List columnLabels = new ArrayList<String>();
+    List<Integer> types = Lists.newArrayList();
+    List<Integer> nullability = Lists.newArrayList();
+    try {
+      int columnCount = resultSet.getMetaData().getColumnCount();
+      for (int i = 1; i <= columnCount; i++) {
+        columnLabels.add(resultSet.getMetaData().getColumnLabel(i));
+        types.add(resultSet.getMetaData().getColumnType(i));
+        nullability.add(resultSet.getMetaData().isNullable(i));
+      }
+      
+      String msg = "\n" + query + "\n" + columnLabels + "\n"
+    		  + this.columnLabels + "\n"
+    		  + "\n" + Utils.getTypesInStrings(types) + "\n"
+    		  + Utils.getTypesInStrings(this.types) + "\n"
+    		  + "\n" + nullability + "\n"
+    		  + this.nullability + "\n";
+      writer.append(msg);
+      
+      if (!columnLabels.equals(this.columnLabels) || !types.equals(this.types)
+    		  || !nullability.equals(this.nullability)) {
+        LOG.info(msg);
+        setTestStatus(TestStatus.VERIFICATION_FAILURE);
+        exception = exception == null? new VerificationException(msg)
+        	: new VerificationException(exception + "\n" + msg);
+      }
+    } catch (IllegalArgumentException | IllegalAccessException e1) {
+      LOG.warn(e1);
+    } finally {
+      if (resultSet != null) {
+        resultSet.close();
+      }
+      if (writer != null) {
+        writer.close();
+      }
+    }
+  }
+  
   @Override
   public void cancel() {
     thread.interrupt();
@@ -262,7 +315,8 @@ public class DrillTestJdbc implements DrillTest {
   }
   
   public synchronized void setTestStatus(TestStatus status) {
-	if (testStatus == TestStatus.CANCELED) return;
+	if (testStatus == TestStatus.CANCELED || testStatus == TestStatus.VERIFICATION_FAILURE) 
+	  return;
 	testStatus = status;
   }
 	 


framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -90,13 +90,18 @@ public void run() {
...
+      
       testVerifier = new TestVerifier(types, query, columnLabels, matrix.verificationTypes);
       if (query.startsWith("explain") || matrix.verificationTypes.get(0).equalsIgnoreCase("regex")) {
         setTestStatus(testVerifier.verifyTextPlan(matrix.expectedFile, outputFilename));
       } else {
         setTestStatus(testVerifier.verifyResultSet(matrix.expectedFile, outputFilename));
       }
+      
+      if (modeler.type.equalsIgnoreCase("limit 0")) {
+    	  String limitZeroQuery = "select * from (" + query + ") limit 0";
```
sudheeshkatkam: Add a table alias:

```"SELECT * FROM (" + query + ") T LIMIT 0"```

File: framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java [hunk 2]
```diff
@@ -241,6 +245,73 @@ private void executeQuery(String query) throws IOException, SQLException {
     }
   }
 
+  private void executeLimitZeroQuery(String query) throws IOException, SQLException {
+	BufferedWriter writer = new BufferedWriter(new FileWriter(new File(outputFilename),true));
+    final boolean cancelQuery = rand.nextInt(100) < TestDriver.OPTIONS.cancelPercent;
```
sudheeshkatkam: Throwing it out there.. Can we ignore the cancel flag for this?

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -241,6 +245,73 @@ private void executeQuery(String query) throws IOException, SQLException {
...
+            return;
+          }
+        } catch (InterruptedException e) {
+          resultSetA.close();
+          writer.close();
+          return;
+        }
+      }
+    }
+    List columnLabelsA = new ArrayList<String>();
```
sudheeshkatkam: You can use [variable shadowing](http://www.xyzws.com/javafaq/what-is-variable-hiding-and-shadowing/15), and name these local variables as is (resultSet, columnLabels, types, ...). To access the instance level variables use *this* (this.resultSet, this.columnLabels, ...).

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -241,6 +245,73 @@ private void executeQuery(String query) throws IOException, SQLException {
...
+    List<Integer> typesA = Lists.newArrayList();
+    try {
+      int columnCount = resultSetA.getMetaData().getColumnCount();
+      for (int i = 1; i <= columnCount; i++) {
```
sudheeshkatkam: Collapse two loops into one?

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -241,6 +245,73 @@ private void executeQuery(String query) throws IOException, SQLException {
...
+      }
+      for (int i = 1; i <= columnCount; i++) {
+        typesA.add(resultSetA.getMetaData().getColumnType(i));
+      }
+      String msg = "\n" + query + "\n" + columnLabelsA.toString() + "\n"
+    		  + columnLabels.toString() + "\n"
+    		  + "\n" + Utils.getTypesInStrings(typesA).toString() + "\n"
+    		  + Utils.getTypesInStrings(types).toString() + "\n";
+      writer.append(msg);
+      if (!columnLabelsA.equals(columnLabels) || !typesA.equals(types)) {
```
sudheeshkatkam: How is type nullability checked?

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -241,6 +245,73 @@ private void executeQuery(String query) throws IOException, SQLException {
...
+        	: new VerificationException(exception.toString() + "\n" + msg);
+      }
+    } catch (IllegalArgumentException | IllegalAccessException e1) {
+      if (testStatus != TestStatus.CANCELED) LOG.warn(e1);
+    } finally {
+      if (resultSetA != null) {
+        resultSetA.close();
+      }
+      if (writer != null) {
+        writer.close();
```
sudheeshkatkam: To confirm, the output file is deleted if the test succeeds, right?

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -252,7 +252,7 @@ public int runTests() throws Exception {
       for (DrillTestJdbc test : verificationFailures) {
         LOG.info(test.getInputFile());
         LOG.info("Query: \n" + test.getQuery());
-        LOG.info(test.getException().getMessage());
+        LOG.info(test.getException() == null? "":test.getException().getMessage());
```
sudheeshkatkam: When is this true? Printing empty string might confuse us.

sudheeshkatkam:
I have a few comments. Looks good overall.

sudheeshkatkam:
Can you squash the commits into one, please?

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: Yes, we don't need to cancel the limit 0 query. Added a check to exit early if original test is canceled.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: done.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: done

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: Not checked. Not sure that's possible. Will investigate.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: By default yes. Will not delete if ran with -w option.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
cchang738: This is needed because it will cause NPE if not checked.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: added nullability check!

cchang738:
new pre-commit test running here http://10.10.104.91:8080/job/Framework%20Regression/235/console

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -52,6 +52,7 @@
   private TestMatrix matrix;
   private Thread thread;
   private List<Integer> types;
+  private List<Integer> nullability;
```
sudheeshkatkam: How about nullabilityStatuses?

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -241,6 +246,54 @@ private void executeQuery(String query) throws IOException, SQLException {
     }
   }
 
+  private void executeLimitZeroQuery(String query) throws IOException, SQLException {
+	if (getTestStatus() == TestStatus.CANCELED) {
```
sudheeshkatkam: Should we run the LIMIT 0 query only if the test status is PASS? This will avoid dealing with multiple errors.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -241,6 +246,54 @@ private void executeQuery(String query) throws IOException, SQLException {
...
+        columnLabels.add(resultSet.getMetaData().getColumnLabel(i));
+        types.add(resultSet.getMetaData().getColumnType(i));
+        nullability.add(resultSet.getMetaData().isNullable(i));
+      }
+      
+      String msg = "\n" + query + "\n" + columnLabels.toString() + "\n"
+    		  + this.columnLabels.toString() + "\n"
+    		  + "\n" + Utils.getTypesInStrings(types).toString() + "\n"
+    		  + Utils.getTypesInStrings(this.types).toString() + "\n"
+    		  + "\n" + nullability.toString() + "\n"
```
sudheeshkatkam: Please add `Utils.getNullabilityStatusesInStrings(...)` for better readability of statuses. The status names are listed in the `java.sql.ResultSetMetaData` class.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -241,6 +246,54 @@ private void executeQuery(String query) throws IOException, SQLException {
...
+      
+      if (!columnLabels.equals(this.columnLabels) || !types.equals(this.types)
+    		  || !nullability.equals(this.nullability)) {
+        LOG.info(msg);
+        setTestStatus(TestStatus.VERIFICATION_FAILURE);
+        exception = exception == null? new VerificationException(msg)
+        	: new VerificationException(exception.toString() + "\n" + msg);
+      }
+    } catch (IllegalArgumentException | IllegalAccessException e1) {
+      if (testStatus != TestStatus.CANCELED) LOG.warn(e1);
```
sudheeshkatkam: Unnecessary condition, just log.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -262,7 +315,8 @@ public void run() {
   }
   
   public synchronized void setTestStatus(TestStatus status) {
-	if (testStatus == TestStatus.CANCELED) return;
+	if (testStatus == TestStatus.CANCELED || testStatus == TestStatus.VERIFICATION_FAILURE)
```
sudheeshkatkam: Is `|| testStatus == TestStatus.VERIFICATION_FAILURE` necessary?

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
sudheeshkatkam: (Better question) We need to ensure that test.getException() does not return null when there is a verification failure, so the cause is known. Is there a case where `test.getException() == null` is true?

sudheeshkatkam:
Please edit the commit message.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -241,6 +246,54 @@ private void executeQuery(String query) throws IOException, SQLException {
...
+    List<Integer> nullability = Lists.newArrayList();
+    try {
+      int columnCount = resultSet.getMetaData().getColumnCount();
+      for (int i = 1; i <= columnCount; i++) {
+        columnLabels.add(resultSet.getMetaData().getColumnLabel(i));
+        types.add(resultSet.getMetaData().getColumnType(i));
+        nullability.add(resultSet.getMetaData().isNullable(i));
+      }
+      
+      String msg = "\n" + query + "\n" + columnLabels.toString() + "\n"
```
sudheeshkatkam: unnecessary `toString()` calls.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
cchang738: It was true for type verification. I have since changed it to also throw a verification exception. Let me revert it back so we always enforce a message.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: Are you suggesting we track types and nullability differently? Currently they are lumped together. I don't feel it's necessary to separate them out. It's very clear by looking at the output what is failing the test.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: There is no harm (in fact preferred) to catch both failures. So we should run limit 0 query. We should not run limit 0 if there is an execution failure. Let me guard on that as well.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: Let's get this merged first. I will file a separate MD for the improvement.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: done

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: Yes, necessary. We may have other verification try to set it to success. It's a good defensive coding.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: Not wrong but removed.

cchang738:
new pre-commit run at http://10.10.104.91:8080/job/Framework%20Regression/236/console

cchang738:
new pre-commit run at http://10.10.104.91:8080/job/Framework%20Regression/237/console

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
sudheeshkatkam: I meant rename this particular variable from nullability to nullabilityStauses.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
sudheeshkatkam: There isn't anything that does that, right? Data verification happens beforehand, so it seems unnecessary. 

Also in execLimitZeroQuery, there should be a throw new VerificationException(...) rather than set status and exception directly. The run method takes care of failing the test.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
sudheeshkatkam: The way the test is setup ( testVerifier.verify... throws an exception that is caught), execLimitZeroQuery will be skipped anyway.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: nullabilityStauses would be too long. The original name is just fine.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
cchang738: You are right for the current setup. But it doesn't prevent one day others might change the framework and make the checking necessary. In any event, this is another defensive coding.

cchang738:
Thanks for review. Pre-commit test was successful. Merge it because we are waiting for this feature.

PR #65, opened 2015-12-09
hive metadata changes

rchallapalli:
Khurram/Krystal,

I added some tests around changing hive metadata. I will still be adding more tests in the future. I got a clean functional run

Rahul


rchallapalli:
I also made a change to bring back the debug statements during the data generation stage

kfaraaz:
Rahul,

verified commit ID of your commit and the clean functional run, both are same.
Tests look good.
LGTM!

Thanks,
Khurram

krystaln:
I checked the tests and they look good to me.

Thanks,
Krystal

PR #67, opened 2015-12-11
Moved tpcds query30.sql to failing - DRILL-4095

vmarkman:
This is just one change: moved query30.sql to failing, DRILL-4095


vmarkman:
Chun, can you please review ? Thank you.

PR #70, opened 2015-12-15
Modify baselines for DRILL-4194

vkorukanti:
DRILL-4194 changed the expected output for HiveScan. Changed the baselines accordingly. Running a regression test: http://10.10.104.91:8080/job/Framework%20Regression/273/


Agirish:
+1 from me, but I think Rahul should take a look at this.

rchallapalli:
There are some failures with the modified tests. I think there are still some issues with escaping

vkorukanti:
Here is the successful run. Had to modify the pattern for escaping some special characters.

rchallapalli:
Thanks venki

PR #71, opened 2015-12-18
Verify

rchallapalli:
Khurram,

I verified a few more bugs and added test cases for them. Below is the functional run URL. The failures are known failures. Can you take a look?

http://10.10.104.91:8080/job/Framework%20Regression/290/console

Rahul


Agirish:
Rahul, could you update the addremove_files.sh file to have an if/else structure to check if directory exists before you create or delete them?

Agirish:
Other than that, looks good.

rchallapalli:
Abhishek,

Made the suggested change and the functional run did not have any issues. Can you take a look?

Rahul

Agirish:
LGTM, +1

PR #72, opened 2015-12-21
Modify baseline in explain plan output for DRILL-4201.

jinfengni:
The order of the clauses in the conjunctive form is switched, due to change made in DRILL-4201.


rchallapalli:
Jinfeng,

Your changes look good. I just kicked off a regression run just to be sure. Once that succeeds, we can merge this change

Rahul

rchallapalli:
Thanks for fixing this jinfeng!

PR #74, opened 2015-12-22
Adding retry attempts for data generation scripts

agirish:
Chun, 

Can you please take a look?


Agirish:
Chun, as discussed in person, I made the changes you suggested. Will also have someone else take a look at it and then merge it.

PR #76, opened 2016-01-06
Changed the hive ddl(orc) so that it runs on hive-13 as well

rchallapalli:
Abhishek,

I modified the DDL around ORC so that the hive queries also run against hive-13. I got a clean functional run. Can you review the change?
- Rahul


Agirish:
Looks good. +1

PR #80, opened 2016-01-12
Moved the failing tests

rchallapalli:
Abhishek,

I moved the 4 known failing hive tests so that they are not run during the functional regression. I scheduled a regression run. Can you review?
- Rahul


Agirish:
Looks good. +1

PR #85, opened 2016-01-21
Fixed a dependency in the ddl script

rchallapalli:
Abhishek,

Make a change to the hive script so that there is no dependency. Can you kindly review it?

Rahul


Agirish:
Looks good. +1

PR #86, opened 2016-01-23
Fixed few more hive dependency issues

rchallapalli:
Abhishek,

Fixed a few more dependencies with the hive ddl's. I will schedule a functional run. Can you review them in the mean time.

Rahul


Agirish:
Looks good. +1

I scheduled a run with Hive 0.13 and that went thru fine.

PR #96, opened 2016-02-05
Test to verify fix for DRILL-4349

khurram-faraaz:
Hello Vicky,

I have a clean run. Changes include a test case to verify fix for DRILL-4349
I have shared the link to clean run in another email.
Please review changes and merge into master.

Thanks,
Khurram


kfaraaz:
Thanks Vicky. Agreed we should definitely add more tests around that area. Thanks for your review and merge.

PR #100, opened 2016-02-13
hive native scan tests

rchallapalli:
Chun,

I added tests around the hive native scan functionality. This is still work in progress (partition pruning tests are pending). Kindly do a review and let me know your thoughts. Functional run completed fine.

Rahul


rchallapalli:
Chun,

For (1), I added comments and also kicked off a regression job just to be sure.
For (2), While I agree we should be able to globally enable/disable hive native reader, that task entails spending more time on the framework to support such behavior. Until we achieve that, I still think having these tests in functional suite adds value. Otherwise we have to manually run them for every release (if not for every commit). Let me know if you think otherwise.

Rahul

cchang738:
Let's collapse these commits into one. It's cleaner. Let's do that for future PR, one commit per PR. Thanks.

rchallapalli:
Collapsed the commits into a single commit and kicked off the regression run. I will commit the change once the run completes

While we want to maintain a cleaner commit history, one PR per commit model is not a sustainable model. Eg : When we are verifying bugs we usually have 1 commit per bug and then run a regression job after verifying a few bugs. Running the regression job for every testcase (corresponding to a bug) is a waste of time and resources

rchallapalli:
Got a clean run after collapsing the commits.

cchang738:
Your comments to one commit needs run a regression job is incorrect. One PR per commit model is not what you think it is. Basically you can have many commits while working, either changing framework source code or verifying bugs and adding tests. But before your create a PR, you should squash the commits into one, then run regression and create the PR, just like what you did for this PR after my comments. Hope this makes sense to you. 

    On Monday, February 15, 2016 4:43 PM, Rahul Challapalli <notifications@github.com> wrote:
 

 Collapsed the commits into a single commit and kicked off the regression run. I will commit the change once the run completesWhile we want to maintain a cleaner commit history, one PR per commit model is not a sustainable model. Eg : When we are verifying bugs we usually have 1 commit per bug and then run a regression job after verifying a few bugs. Running the regression job for every testcase (corresponding to a bug) is a waste of time and resources—

rchallapalli:
Got it. Thanks!

PR #101, opened 2016-02-16
Partition pruning tests around hive when using drill's native parquet reader

rchallapalli:
Chun,

I added tests around partition pruning and got a clean functional run. Can you kindly review the changes?

Rahul


cchang738:
Are they existing tests that are wrapped around with the alter statement? If so, I don't think we need to add them as new tests. The automation framework already support global settings. We just need to run them with native scan enabled. 

If they are new tests, let's remove the alter statement. Let's restructure our tests and configure test runs with native scan enabled. The thing is if we don't do it now, we probably will never do it the right way.

rchallapalli:
I selected a sub-set of existing tests and also added plan verification. 

Your suggestion makes sense. We should configure a new test job with native scan enabled and run all the hive tests. 
However, I want at-least some sanity tests around partition pruning (with native scan) to be part of the pre-commit run so that we catch things sooner rather than later.

cchang738:
In that case, +1.

PR #102, opened 2016-02-18
Verified drill4250 and added a testcase

rchallapalli:
Abhishek,

I verified drill-4250 and added a testcase. Can you review and let me know your comments

Rahul


Agirish:
Looks good. +1

Just to confirm, these are tests added to an existing suite and the Datasources copy would happen as a part of it.

rchallapalli:
Thanks Abhishek. You are right, the newly added folders with automatically be copied.

PR #104, opened 2016-02-25
MD-691: Pre test check

cchang738:
For now we only check number of cluster nodes. That is the number of cluster nodes should be equal to number of drill bits.

http://10.10.104.91:8080/job/Framework-Regression/459/console

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/CmdConsOut.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/CmdConsOut.java
@@ -1,0 +1,7 @@
+package org.apache.drill.test.framework;
+
+public class CmdConsOut{
+  public int exitCode;
+  public String consoleOut;
+  public String cmd;
+}
--- a/framework/src/main/java/org/apache/drill/test/framework/ConnectionPool.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/ConnectionPool.java
@@ -36,8 +36,14 @@ public class ConnectionPool implements AutoCloseable {
 
   private final Map<String, Queue<Connection>> connections;
 
-  public ConnectionPool() throws ClassNotFoundException {
-    Class.forName("org.apache.drill.jdbc.Driver");
+  public ConnectionPool() {
+    try {
+		Class.forName("org.apache.drill.jdbc.Driver");
+	} catch (ClassNotFoundException e) {
+		// TODO Auto-generated catch block
+		e.printStackTrace();
+		System.exit(-1);
+	}
     Driver.load();
     connections = new HashMap<>();
   }
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -59,6 +59,7 @@ public class TestDriver {
   private String drillTestData = drillProperties.get("DRILL_TESTDATA");
   private String fsMode = drillProperties.get("FS_MODE");
   private Connection connection = null;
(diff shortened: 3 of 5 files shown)


rchallapalli:
1. There should only be one ConnectionPool instance throughout. You seem to be instantiating the ConnectionPool class in the static block of Utils class which seems odd. I would rather make ConnectionPool class a singleton.

2. I would consider creating a ClusterInfo class. Each attribute will be a cluster property which we want to verify. This will make the code clean. Instead of having each cluster property as an attribute, you might also want to think about creating a Pair <attribute, validating_query>. This way we can add a validate() method to the ClusterInfo class which will run the corresponding query for each attribute and returns true if the output of the query matches the attribute value for all such pairs.

cchang738:
Thanks for reviewing.

1. Currently, we only use one instance of the connectionPool. But that does not mean we may not have multiple connectionPools in the future. For example, I may very much likely creating a new connectionPool for testing client impersonation. Leave it this way is more flexible.

2. The cluster info class is a good idea. I will consider it when we have other things to check. For now, let's get this in and I will leave MD-691 open for input (we need to discuss what are the things worth checking).

rchallapalli:
1. Unless we need to create connections with specific properties, I do not see why we need multiple instances of ConnectionPool. May be that's the case with client impersonation. Even then instantiating in the Utils class is weird. I would rather create a Factory for connections which returns the right connection based on the desired properties and make it a static variable of the TestDriver class.

2. From our discussions, adding more properties is very likely. I would rather fix it now than later. We grossly under-estimate the time that hacks actually live in the code.

cchang738:
Why is it weird? Utils is where you provide utility functions to do things or access resources, in this case, connectionPool. What if down the road (and likely) we need to access the connectionPool from another class other than TestDriver? The logic way to handle that is to go to your tool box and ask for it.

For creating the ClusterInfo class, I will spend more time on it. Things I can think of now, include IP addresses, Drill version. These information already captured currently. So no real urgency.

rchallapalli:
We can argue about whether Utils is the right place instantiating ConnectionPool, but that is not the main point. My main point is that we should only instantiate ConnectionPool once (make it a singleton). If you think you need a different connection object based on specific properties, then just change the getOrCreateConnection method to accept a "Properties" argument

For the ClusterInfo class lets start with "no of drillbits" property first. We will add other things later

cchang738:
Rahul, that's the other area that I can't quite understand your logic. Why do you want to artificially limit the connectionPool to a singleton? The test framework is not a server application where you should maintain a single client connection pool. Here as a client application, we have many scenarios that we need to have multiple connection pools. I can give you another example. Drill supports accessing external DB as data source. Let's say we have tests that need to connect to the external DB and making constant changes while we also accessing DB data through Drill. It's a lot easier to manage these separate connection end points with separate connection pools. The point is, arguing we must have a singleton here is, the best not very productive, the least shortsighted.

rchallapalli:
As long as we are talking about connections to Drill, I do not see an issue with making ConnectionPool a singleton. Now why do we need a singleton? The whole point of using a ConnectionPool is to reuse connections. By creating a different instance, you are creating an entirely new pool of connections. This is a waste of resources. Though Steven did not make the class a singleton, he made sure it is only used instantiated once by passing references whenever necessary

Now if you intend the framework to talk with other sources directly (external DB) and not through drill, then it is a completely different use case and I would lean towards a factory class which returns the appropriate connection pool for that data source. Even here we should only have single instance of a connection pool per data source.

cchang738:
clean up PR for more framework change.

PR #105, opened 2016-02-25
Fixed issues with metadata_caching suite related to the RPM

rchallapalli:
Abhishek,

I modified how the metadata suite is copied to the maprfs cluster and changed the folder permissions so that the Drillbit process user owns the entire folder. I got a clean regression run. However this change is currently not compatible with the Docker setup. We can discuss whether it is even a possible solution to update the docker setup to support passwordless ssh from the machine which runs the test framework onto the drill cluster.

Rahul

### DIFF
--- a/framework/resources/Datasources/metadata_caching/copy_data.sh
+++ b/framework/resources/Datasources/metadata_caching/copy_data.sh
@@ -1,0 +1,10 @@
+source ~/.drillTestConfig
+if ! $(hadoop fs -test -d /drill/testdata/metadata_caching); 
+then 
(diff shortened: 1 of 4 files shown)


Agirish:
Looks good. To find the Drillbit user, can't we make use of some (possibly new) variable in [drillTestConfig](https://github.com/mapr/drill-test-framework/blob/master/framework/src/main/resources/drillTestConfig)
That might help make it simple and work on Docker as well.

kfaraaz:
If any new variable is added to drillTestConfig please share the details with the team, thanks!

PR #108, opened 2016-03-04
Remove trivial projects from tests

jacques-n:
Correct casts to be explicit varchar(1) where necessary.
Correct order of columns in hbase projections.


cchang738:
Was a regression run against the PR? If yes, please post the summary. Thanks.

jacques-n:
No, it was not. We cannot run this suite since it is MapR specific.

cchang738:
OK. I will try at MapR end.

PR #109, opened 2016-03-05
Hive Tpcds SF100 hive generated parquet

rchallapalli:
Abhishek,

I modified the tpcds sf100 parquet hive suite so that we do not generate the required data using map-reduce jobs on the fly. I had a clean run on the Advanced suite. I will share the location of the hive generated data so that you can make it work with docker. This should enable us to run the Advanced suite on docker. Kindly review the change

Rahul


Agirish:
Rahul, I'm taking a look. Sorry for the delay - was held up with something.

Agirish:
Looks like the scripts need to be updated to specify the location of the src tables?

PR #112, opened 2016-03-14
New tests to cover frame clause support in window definitions

khurram-faraaz:
Hello Vicky,

I have these tests that cover the new frame clause support in window definitions. 
Please review them and if they look good, please merge the changes to master.
I have full functional run, all my new tests pass. There are some unrelated failures which are not caused by my tests and they are not tests related to frame clause.

Thanks,
Khurram


vmarkman:
Looks good. +1

kfaraaz:
Thanks Vicky for your review, there are two different JIRA numbers. I will add that info as part of my next commit.

PR #121, opened 2016-03-31
MD-839: we should not explicitly test hash64(). remove the test

cchang738:
@krystaln @kfaraaz can one of you please review the PR? it's just remove this one test. we should not explicitly test hash64() as decided in one of the standup.

http://10.10.104.91:8080/job/Framework-Regression-Advanced/80/console


krystaln:
LGTM +1

PR #123, opened 2016-04-07
fixed few hive dependencies

rchallapalli:
Abhishek,

I fixed hive dependencies in half of the suites and I got a clean functional run. Can you take a look?

Rahul


Agirish:
Looks good. +1

PR #124, opened 2016-04-07
Fixed more hive dependencies

rchallapalli:
Abhishek,

With this set of changes, we should be good to run the hive suites on docker. Can you kindly review these changes?

Rahul


Agirish:
Looks good. +1. 

Thanks for making these changes.

PR #127, opened 2016-04-14
MD-360: implement cancelling query after timeout

cchang738:
@sudheeshkatkam 

Can you review the PR as we discussed? I checked, after statement.cancel(), no other call can be made to check status.
Thanks.
http://10.10.104.91:8080/job/Framework-Regression/508/console

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
@@ -36,6 +36,7 @@ import java.sql.Types;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class DrillTestJdbc implements DrillTest {
   private static final Logger LOG = Logger.getLogger(DrillTestJdbc.class);
@@ -55,6 +56,8 @@ public class DrillTestJdbc implements DrillTest {
   private List<Integer> columnNullabilities;
   private List columnLabels = new ArrayList<String>();
   private Random rand = new Random();
+  private Statement statement = null;
+  private AtomicBoolean doneProcessingResultSet = new AtomicBoolean(false);
 
   public DrillTestJdbc(DrillTestCase modeler, ConnectionPool connectionPool) {
     this.modeler = modeler;
@@ -143,7 +146,7 @@ public class DrillTestJdbc implements DrillTest {
       if (query.contains("alter")) {
         LOG.debug(query + " " + connection.hashCode());
       }
-      Statement statement = connection.createStatement();
+      statement = connection.createStatement();
       resultSet = statement.executeQuery(query);
     } finally {
       if (resultSet != null) {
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -300,6 +301,13 @@ private boolean isNullabilityCompatible(List<Integer> limitZero, List<Integer> r
...
   @Override
   public void cancel() {
+	if (statement != null)
+		try {
+			statement.cancel();
+		} catch (SQLException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
+		}
     thread.interrupt();
```
sudheeshkatkam: If the query was cancelled successfully, the `thread` will be unblocked in near future. So this `interrupt()` call could "hide" incorrect handling of cancellation.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -226,6 +228,7 @@ private void executeQuery(String query) throws IOException, SQLException {
           writer.write(columnList + "\n");
         }
       }
+      doneProcessingResultSet = true;
```
sudheeshkatkam: Use an AtomicBoolean, and move this statement to finally block.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -300,8 +303,30 @@ private boolean isNullabilityCompatible(List<Integer> limitZero, List<Integer> r
...
+		e.printStackTrace();
+		return;
+	  }
+	  i++;
+	}
+	if (!doneProcessingResultSet) {
+      
+	  LOG.warn("Cancel after timeout may have failured!");
+	}
+	thread.interrupt();
```
sudheeshkatkam: Statement should be in the if condition block.

sudheeshkatkam:
Comments to be addressed. Otherwise, +1

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -300,8 +303,30 @@ private boolean isNullabilityCompatible(List<Integer> limitZero, List<Integer> r
...
+		thread.sleep(1000);
+	  } catch (InterruptedException e) {
+		e.printStackTrace();
+		return;
+	  }
+	  i++;
+	}
+	if (!doneProcessingResultSet) {
+      
+	  LOG.warn("Cancel after timeout may have failured!");
```
sudheeshkatkam: failed*

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -300,8 +305,30 @@ private boolean isNullabilityCompatible(List<Integer> limitZero, List<Integer> r
...
+		statement.cancel();
+	  } catch (SQLException e) {
+		LOG.warn("Cancel after timeout failed!");
+		e.printStackTrace();
+	  }
+	}
+	int i = 0;
+	while (!doneProcessingResultSet.get() && i < 10) {
+	  try {
+		thread.sleep(1000);
```
sudheeshkatkam: Current thread should sleep, not `thread`.

cchang738: Good catch. Thanks.

PR #133, opened 2016-04-26
Running the same test multiple times

rchallapalli:
Chun/Abhishek,

I added the ability where we can run the same test multiple times. I also added a suite for network intensive queries. Let me know your thoughts.
- Rahul


Agirish:
Hey Rahul, 

I think it would be good to separate the test cases from framework code. Should we make them as independent commits?

rchallapalli:
Sure Abhishek. I can do that

cchang738:
Rahul,

I think we can implement this feature in a much simpler way. No need to create new class and interface. Just a very simple change to existing DrillTestJdbc.java and TestDriver.java. I will discuss this with you in person.

Chun

rchallapalli:
Folks,

I am closing this PR. I will create a new PR which has different commits for the framework changes and the tests themselves

PR #137, opened 2016-04-28
MD-877: clone tests to run multiple times concurrently

cchang738:
@rchallapalli Please review.

Had regular run. Also ran with the -j option. See below link.
http://10.10.104.91:8080/job/Framework-Regression/538/console


rchallapalli:
LGTM +1

PR #153, opened 2016-05-30
date_trunc function tests for review

khurram-faraaz:
Hi Rahul,

Can you please review these new tests for date_trunc function and merge them after your review.
I have disabled three tests (for now) the ones that failed due to difference in timezone handling.
I have a clean run, for which I will share the link.

Thanks,
Khurram


rchallapalli:
Khurram,

Thanks for adding these tests Khurram. Can you also cover the below cases ?
  1. Read values from a file. Also can you make sure we have nulls in the file. A few cases around this
  2. Can you have atleast one query where we use date_trunc function in "where", "group by", and "join" operators. Also something like "max(date_trunc(.....))" would be nice to have

Rahul

kfaraaz:
Sure Rahul. Thanks for the review and comments. I will definitely add the new scenarios that you have mentioned in another PR. If these tests look good please merge.
I will come up with new tests for the areas that you have listed in your comments.

Regards,
Khurram

kfaraaz:
Thanks Rahul.

PR #155, opened 2016-06-06
Diabled date_trunc tests

khurram-faraaz:
Hi Rahul,

I have disabled date_trunc tests as there were some failures due to timezone handling.
I have also disabled one values test that fails only when assertions are enabled. values/values_int_8.q
Please review and merge

Thanks,
Khurram


rchallapalli:
+1

PR #163, opened 2016-07-01
Adding tests to p1tests for fix to DRILL4707

rhou1:
: Conflicting columns names under case-insensitive policy lead to either memory leak or incorrect result.

These tests use column names where the cases of the letters have been changed in various ways.

Chun or Rahul, can you please review the change?

Thanks.

--Robert


rchallapalli:
The tests look good. Few comments 

    1. You added some hive tests under p1tests folder. Ideally we would want hive tests to be in Functional/hive/hive_storage or atleast in their own folder. So that we can skip running them if desired.
    2. For every suite which relies on hive, the test definition file should mention the dependency on hive. You can refer to https://github.com/mapr/drill-test-framework/blob/master/framework/resources/Functional/hive/hive_functions/hive_functions.json

rchallapalli:
LGTM +1

PR #166, opened 2016-07-06
Surround hive's IF function with backticks

rchallapalli:
Krystal,

Once drill-4673 get committed, "IF" will become a keyword in drill. So I modified the hive tests which use "IF" by escaping the keyword. Can you review this change?

Rahul


krystaln:
Looks good. +1

rchallapalli:
Thanks Krystal

PR #170, opened 2016-07-13
Resolved issues with multiple suites having the same name

rchallapalli:
Chun,

I fixed the issues with all suites having test-id conflicts in Functional. Can you take a look? I had a clean functional run. However I see that you merged a change recently. So if everything looks good, I will rebase and run before merging them. 

Rahul


cchang738:
I saw some file changed to a different name, like this one

...rge-join/nullable-parquet-hive-join.e_tsv → ...-join/mj_nullable-parquet-json-join.e_tsv

Is this correct? If not, can you go over the others and make sure they are renamed correctly. 

I don't think you need to rebase and rerun. But it's your choice.

rchallapalli:
Reason for renames : We have a sub-folder which uses the test definition file from the parent folder.There could be a conflict when the testcases in the parent folder and the subfolder have the same name (since they share the same test id as well).

cchang738:
I know that. I was referring to nullable-parquet-hive-join.e_tsv was renamed to mj_nullable-parquet-json-join.e_tsv, from hive-join.e_tsv to json-join.e_tsv. Feels wrong.

rchallapalli:
Hmm....I think git's diff tool just got confused since I renamed multiple files in the same directory. Take a look at the actual folder at [1]. You can see  that all the files are present

[1] https://github.com/rchallapalli/drill-test-framework/tree/MD982/framework/resources/Functional/cross-sources/merge-join

cchang738:
If you have multiple files that got moved in the same PR and the contents of these files are the same, then git may make wrong name connection. This is not a problem. But please do verify the contents are the same for those mismatched. If they are, +1.

rchallapalli:
I also verified that some of the files which github suggests could have been renamed improperly have the same exact content. This leads to the same hash being generated for the 2 files and hence the confusion

rchallapalli:
Thanks Chun!

PR #171, opened 2016-07-18
Verified drill4665 and added testcases

rchallapalli:
Krystal,

These are the tests for drill-4665 where partition pruning failed when we have "like" on a non-partitioned column. Kindly review the testcases added.

Rahul


krystaln:
LGTM +1

rchallapalli:
Thanks Krystal

PR #178, opened 2016-07-30
Added reporting support

prasadns14:
-Added support to log the test case execution summary to a json file
-Each test case execution summary has the following info - drill version, commitID, query, status, queryPath, exception(if any)
- Reporting is disabled by default, use the -r option to enable

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -198,7 +198,7 @@ public class TestDriver {
         }else{
           document.set("errorMessage", "N/A");
         }
-        document.set("runtime", test.getDuration().toString());
+        document.set("queryExecutionTime", test.getDuration().toString());
         document.set("drillVersion", version);
         document.set("commitId", commitId);
         bufferedWriter.write(document.toString());


Agirish:
This is great - thanks for making this change. The regression run is successful too. I've taken a brief look. Before I give a +1, I want to take another look tomorrow. Also let's have Chun review it once.

cchang738:
@prasadns14 Great you are interested in the testing code and put in effort to enhance it. I just took a brief look. Have some questions and/or maybe suggestions. 

1. Can you comment on what's the use case? How will this enhancement be used? I know it's turned off by default, but in what situation we want to turn it on?

2. To make this more useful, I am thinking maybe in addition to what you have, we can include some running stats, such as  test duration. 

3. How is testId defined/initialized? There is an Id assigned for each test in the existing code. That was added to accommodate running one test multiple times. Can you look into that and see if there is any conflict or opportunity to combine them.

Agirish:
Chun, this feature is a requirement from my side. To give you the whole picture, once we have this feature in, going forward, when we enable this for say the nightly runs, it creates and saves a query-able report. We can then run a query like "select * from reports where queryID=<> and result=FAIL, to find regressions and more. 

We could capture query stats and any more information everyone suggests - but I'm thinking we could move it to a next checkin. 

Regarding testID, we made it a combination of test-suite-ID + query name. But we must look into what you said.

prasadns14:
Hi Chun,
Answering your queries-
1) Reporting will help us have a history of the test runs. These logs can be stored and queried in the future to get more statistics
2) Sure, I can include query runtime statistic. I can add more statistics as needed
3) I have to look into the code to figure out if there is any conflict with testId. Will update once i am done.

prasadns14:
-Updated the id to accommodate cloneId and iteration number. The id generated is of the following format - <testId>_<queryName>_<cloneId>_<iterationNumber>
-Added test duration stats

Agirish:
Prasad, please close this pull request. Rebase your branch on master, rerun tests and create a new pull request, which can then be reviewed and merged. Let me know if you need help

PR #182, opened 2016-08-09
Support for json reports

prasadns14:
- Added support to generate json reports
- Regression run passed


Agirish:
+1

Agirish:
Thanks for making this change Prasad.

PR #184, opened 2016-08-10
Tests for nested aggregates support in window functions.

khurram-faraaz:
Hello Rahul,

I have removed the .swp file and I have a clean run of my tests. I will share the link to pre commit test run.

Summary of tests added as part of this commit are,
Test support for nested aggregates in window functions, covered the below areas
- test nested aggregates on all supported data types
- test nested aggregates with an empty over clause
- test nested aggregates using different supported frame clauses
- test nested aggregates with multiple window definitions
- test nested aggregates with no ORDER BY in window definition
- test nested aggregates with no PARTITION BY in window definition
- test nested aggregates over views
- test nested aggregates from within a subquery
- test nested aggregates in a CTE query
- negative tests

Thanks,
Khurram

### DIFF
--- a/conf/drillTestConfig.properties.example
+++ b/conf/drillTestConfig.properties.example
@@ -7,7 +7,7 @@ NUMBER_OF_CLUSTER_NODES=4
 ########################################
 # Drill home
 ########################################
-export DRILL_HOME=/opt/drill
+DRILL_HOME=/opt/drill
 
 ########################################
 # Drill test data dir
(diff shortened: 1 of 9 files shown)


Agirish:
Test Suite ID seems unique. And these tests are on existing datasets right? 

Looks good. +1

As discussed, kindly enable the new TPC-DS queries in a later check-in.

kfaraaz:
Yes all new tests use existing parquet data. 
I will enable TPC-DS tests as part of my next commit.

Thanks,
Khurram

PR #187, opened 2016-08-19
Json reports

prasadns14:
Added option to copy reports to DFS


Agirish:
Looks good. Please schedule a regression run and update once done. Will have it merged in.

cchang738:
Looks like you will have two copies of the report, one on local filesystem, and one on DFS. Do we really need two? What if the DRILL_REPORT_DFS_DIR is not defined?

Agirish:
We will need the one on DFS - but I agree it should be optional. We could either include a flag for it , or skip uploading if the var is not defined.

PR #188, opened 2016-08-22
Fix a bug which may lead to hang process  caused by Utils.execCmd(cmd)

beyond1920:
Utils.execCmd(cmd)  hangs sometimes, because both standard error stream and standard output stream of the Process have to be consumed. Otherwise hang will occasionally occurs like ghost. The JDK's Javadoc documentation provides the answer to this question: Because some native platforms only provide limited buffer size for standard input and output streams, failure to promptly write the input stream or read the output stream of the subprocess may cause the subprocess to block, or even deadlock. More details could be found at http://www.javaworld.com/article/2071275/core-java/when-runtime-exec---won-t.html.


cchang738:
@beyond1920 This looks good. Thanks for fixing it. One comment. Possible to store standard output Stream as well, as you did for error Stream? There are cases we want to capture both and store it to cmdConsOut.sonsoleOut.

beyond1920:
ok, i will do that soon.

beyond1920:
Sorry to close the previous pull request accidentally. So open a new pull request #191. Please review the new PR. Thanks very much.

PR #189, opened 2016-08-22
change the compare logic of ColumnList 

beyond1920:
change the compare logic of ColumnList because some cases work unexpectedly, for example (-3.0f, -2.0f) or (-2.0f, 1.0f). in those cases, expect return false, but actually return true.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
@@ -116,7 +116,7 @@ public class ColumnList {
           float f1 = (Float) list1.get(i);
           float f2 = (Float) list2.get(i);
           if ((f1 + f2) / 2 != 0) {
-            if (!(Math.abs(f1 - f2) / ((f1 + f2) / 2) < 1.0E-6)) return false;
+            if (!(Math.abs((f1 - f2) / ((f1 + f2) / 2)) < 1.0E-6)) return false;
           } else if (f1 != 0) {
             return false;
           }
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -116,7 +116,7 @@ private boolean compare(ColumnList o1, ColumnList o2) {
           float f1 = (Float) list1.get(i);
           float f2 = (Float) list2.get(i);
           if ((f1 + f2) / 2 != 0) {
-            if (!(Math.abs(f1 - f2) / ((f1 + f2) / 2) < 1.0E-6)) return false;
+            if (!(Math.abs(Math.abs(f1 - f2) / ((f1 + f2) / 2)) < 1.0E-6)) return false;
```
yufeldman: How about removing the internal Math.abs altogether?

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -125,7 +125,7 @@ private boolean compare(ColumnList o1, ColumnList o2) {
           double d1 = (Double) list1.get(i);
           double d2 = (Double) list2.get(i);
           if ((d1 + d2) / 2 != 0) {
-            if (!(Math.abs(d1 - d2) / ((d1 + d2) / 2) < 1.0E-12)) return false;
+            if (!(Math.abs(Math.abs(d1 - d2) / ((d1 + d2) / 2)) < 1.0E-12)) return false;
```
yufeldman: same as above

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
beyond1920: good idea

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
beyond1920: thanks, done

cchang738:
Yes, the abs should be applied to the whole equation, otherwise, you might end up with a negative value which will compare to true since a negative is always less than a positive. 

LGTM +1

cchang738:
This is merged in another PR. Thanks for the contribution.

PR #191, opened 2016-08-23
Fix a bug which may lead to hang process caused by Utils.execCmd(cmd)

beyond1920:
#188 Sorry to close the previous pull request accidentally. So open a new pull request.

Utils.execCmd(cmd) hangs sometimes, because both standard error stream and standard output stream of the Process have to be consumed. Otherwise hang will occasionally occurs like ghost. The JDK's Javadoc documentation provides the answer to this question: Because some native platforms only provide limited buffer size for standard input and output streams, failure to promptly write the input stream or read the output stream of the subprocess may cause the subprocess to block, or even deadlock. More details could be found at http://www.javaworld.com/article/2071275/core-java/when-runtime-exec---won-t.html.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
@@ -117,7 +117,7 @@ public class ColumnList {
           float f1 = (Float) list1.get(i);
           float f2 = (Float) list2.get(i);
           if ((f1 + f2) / 2 != 0) {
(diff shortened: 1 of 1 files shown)


cchang738:
@beyond1920 I reviewed your changes. Looks really good. Can you squash your changes into one commit and create a new PR please? Include the ColumnList change as well. I will do more testing here and merge after. Thanks so much.

beyond1920:
@cchang738 , I already commit ColumnList change to this PR as well.

cchang738:
Thanks for the contribution @beyond1920. It's merged in another PR.

PR #192, opened 2016-08-23
Added tests to verify DRILL-4474

khurram-faraaz:
Hi Rahul,

I have added tests to verify DRILL-4474, please verify. These are aggregate count queries with use of distinct within count aggregate function. I have a clean pre-commit test run will share details in email.

Thanks,
Khurram


rchallapalli:
Khurram,

The tests look good. You can go ahead and merge them. Since this is a very bad bug, a few more scenarios might be worth testing
  1. Test this functionality on top of a view (created using '*' and another view created using explicit columns). You can play with a few combinations here.
  2. Also do you think that we should add atleast one sanity test with hive, hbase, & csv formats
  3. A test case where we use a similar count(..) function in a filter condition.

Thoughts?

kfaraaz:
Thanks Rahul for your review and comments.
I will add tests for (1.) and (2.) in my next commit, and also include CSV and JSON data in my tests.

PR #197, opened 2016-09-08
MD-1116: columnLabels need to be initialized for each iteration

cchang738:
@Agirish Can you please review the PR? Thanks.


Agirish:
Thanks for fixing this Chun! LGTM. +1

PR #204, opened 2016-10-03
Added parquet limit tests

rchallapalli:
Abhishek,

Based on padma's request, I added some limit tests on top of parquet. In-order to make sure that there is no randomness in the list of records returned, I only used a single parquet file which is smaller than an MFS block. I had a clean functional run (with known failures). Can you review the PR?

Rahul


Agirish:
Looks good. +1

PR #205, opened 2016-10-11
Adding TPC-H (SF1) Tests on Parquet & JSON datasets

agirish:


Agirish:
Rahul, can you please take a look? I"ll share test run results offline

rchallapalli:
It is very confusing to see a lot of irrelevant commits in the PR. Can you take care of such issues in the future? 
Also when you run any regression jobs, please use the specific commit id instead of just specifying the branch name.

LGTM +1

Agirish:
@rchallapalli done - i've rebased my branch on master. Should only list relevant commits here and henceforth.

PR #215, opened 2016-10-20
Updated the metadata cache file to match the parquet date compatibility changes

rchallapalli:
Parth,

I got a clean functional run with this change. Can you review this change?

Rahul


parthchandra:
Looks like you need to update the new metadata cache file to 'wildcard' the hostAffinity section, since the same file can be used as a baseline on another cluster where the affinities may be different.
The changes related to the date are fine in the new file.

parthchandra:
+1. LGTM.

PR #216, opened 2016-10-20
Fixing test result baselines (TPC-H SF1 dataset)

agirish:


cchang738:
Do you understand what caused the baseline change for query25_min.e_tsv?

Agirish:
The results for query 25 have been verified with Greenplum.

cchang738:
OK. Thanks. +1

Agirish:
Not sure. But the new values were verified with Postgres.

PR #218, opened 2016-10-28
MD-1120: added summary for iterations

cchang738:
@Agirish Can you review this PR? I've tested locally as well. Thanks.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -259,16 +259,19 @@ public class TestDriver {
       }
     }
 
+    int totalPassingTest = 0;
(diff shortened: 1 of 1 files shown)


Agirish:
Hey Chun, thanks for making this change. Do we only print the count of tests that passed, failed (& such) or also the list of failures, along with the exception messages for Execution Failures & Baseline mismatch summary per test for Verification Failures? 

Although the total count will be useful, more useful for analysis would be the ones above.

cchang738:
@Agirish All the other messages are logged during each iteration, so they are available if needed to be looked at. If you feel strongly we need to log them again in the final summary, please file a separate JIRA for further discussion. Thanks for the review.

Agirish:
Agreed. +1

PR #230, opened 2016-11-11
Added tests for drill-4203 which corrects the way in which date is written

rchallapalli:
Krystal,

I added tests for DRILL-4203. You can find the test plan at [1]. I ran the functional regression and the advanced suites. Apart from some unrelated issues, everything ran fine. Let me know if you have any questions.

- Rahul


rchallapalli:
Found an issue with one test. I will modify and open a new PR. Thank You.

PR #231, opened 2016-11-11
Added tests for drill-4203 which corrects the way in which date is written

rchallapalli:
Krystal,

I ran the functional and advanced tests successfully. Can you take a look and let me know your thoughts?

Rahul


krystaln:
LGTM +1

PR #232, opened 2016-11-17
Fixing an issue with concurrent file copy

agirish:


Agirish:
Chun, please review. I have a few successful runs (both Functional and Advanced). Have a few more scheduled - will share results with you offline.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -501,9 +502,11 @@ public void cancel() {
           @Override
           public void run() {
             try {
-              Path src = new Path(CWD + "/" + Utils.getDrillTestProperties().get("DRILL_TEST_DATA_DIR"), datasource.src);
+              Path src = new Path(CWD + "/" + Utils.getDrillTestProperties().get("DRILL_TEST_DATA_DIR") + "/" + datasource.src);
               Path dest = new Path(drillTestData, datasource.dest);
-              hdfsCopy(src, dest, false, fsMode);
+
+              LOG.debug("Copy " + src + " to " + dest);
```
cchang738: I think it's better leave the debug log inside the function. If you have multiple calls to the function, you don't need to add multiple LOG lines.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -555,34 +558,32 @@ public void run() {
...
-    FileSystem localFs = FileSystem.getLocal(conf);
-    if (localFs.getFileStatus(src).isDir()) {
-      for (FileStatus file : localFs.listStatus(src)) {
-        Path newSrc = file.getPath();
-        Path newDest = new Path(dest, newSrc.getName());
-        hdfsCopy(file.getPath(), newDest, overWrite, fsMode);
-      }
-    } else if (!fs.exists(dest) || overWrite) {
+
+    if (!fs.exists(dest)) {
```
cchang738: What happens to the "overWrite" case? Have you tested force write?

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
Agirish: Moved it inside.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
Agirish: We've never supported overWrite, it was hardcoded to false. So I just removed it. We can check support for overWrite on a later change.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -501,9 +502,11 @@ public void cancel() {
           @Override
           public void run() {
             try {
-              Path src = new Path(CWD + "/" + Utils.getDrillTestProperties().get("DRILL_TEST_DATA_DIR"), datasource.src);
```
Agirish: No I guess it was just to make it more clear.

Agirish:
I found an issue. Closing this PR. Will reopen one I have a fix.

PR #233, opened 2016-11-18
Fixing an issue with concurrent file copy

agirish:


framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -501,9 +502,9 @@ public void cancel() {
           @Override
           public void run() {
             try {
-              Path src = new Path(CWD + "/" + Utils.getDrillTestProperties().get("DRILL_TEST_DATA_DIR"), datasource.src);
+              Path src = new Path(CWD + "/" + Utils.getDrillTestProperties().get("DRILL_TEST_DATA_DIR") + "/" + datasource.src);
```
cchang738: What is the consideration of making this change?

Agirish: I've already addressed this is in a previous review.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -555,34 +556,39 @@ public void run() {
...
+        for (FileStatus file : localFs.listStatus(src)) {
+          Path srcChild = file.getPath();
+          Path newDest = new Path(dest + "/" + srcChild.getName());
+          dfsCopy(srcChild, newDest, fsMode);
+        }
+      } else {
+        if (!fs.exists(dest.getParent())) {
+          fs.mkdirs(dest.getParent());
+        }
+        FileUtil.copy(localFs, src, fs, dest, false, fs.getConf());
```
cchang738: it would be nice to know the difference between FileUtil.copy and fs.copyFromLocalFile. But if all your tests prove positive, let's check it in. 

+1

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -555,34 +556,39 @@ public void run() {
...
+        LOG.debug("Copying " + src + " to " + dest);
       }
-    } else {
+    } catch (FileAlreadyExistsException e) {
+      LOG.debug("The source file " + src
+              + " already exists in destination.  Skipping the copy.");
+    } catch (IOException e) {
```
cchang738: Do you still think we should and it's safe to ignore the IOException?

Agirish: As a general statement, no. But from my tests, the only instances where I've seen IO Exception occur here was when a file exists. I think we can take that risk. If we happen to encounter any related suspicious issues, we can take another closer look.

Agirish:
Thanks Chun. I'll wait for a few days before I commit this - I wish to test out a few more things.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
Agirish: Apart from what we've discussed in person, I'm curious too, to find out more about their implementation differences. Will let you know what I find out.

PR #234, opened 2016-11-18
Adding support for Drill Reports on DFS

agirish:
If the property DRILL_REPORTS_DFS is set in drillTestConfig.properties, then when run with the -r option to generate reports, the report is copied over to DFS. 

To query the reports, run the following query (default DFS Reports location is /drill/reports):
select * from dfs.drill.reports;

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -180,19 +180,19 @@ public class TestDriver {
         drillReportsDir = CWD;
       }
 
-      File drillReportDir = new File(TestDriver.drillReportsDir);
+      File drillReportDir = new File(drillReportsDir);
       FileSystem localFS = FileSystem.getLocal(conf);
       FileSystem DFS = FileSystem.get(conf);
 
       if (!drillReportDir.exists()) {
         if (!drillReportDir.mkdir()) {
-          LOG.debug("Cannot create directory " + TestDriver.drillReportsDir
+          LOG.debug("Cannot create directory " + drillReportsDir
                   + ".  Using current working directory for drill output");
-          TestDriver.drillReportsDir = CWD;
+          drillReportsDir = CWD;
         }
       }
 
-      File reportFile = new File(TestDriver.drillReportsDir + "/apache-drill-" + version + "_" + commitId + "_" +
+      File reportFile = new File(drillReportsDir + "/apache-drill-" + version + "_" + commitId + "_" +
               "report_" + new Date().toString().replace(' ', '_').replace(':','_') + ".json");
 
       BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(reportFile));


Agirish:
Chun, please review. I've tested report creation with both the variable set and unset. Works as expected.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -55,7 +52,8 @@
   protected static Map<String, String> drillProperties = Utils
       .getDrillTestProperties();
   public static final String drillOutputDirName = drillProperties.get("DRILL_OUTPUT_DIR");
-  public static String drillReportDirName = drillProperties.get("DRILL_REPORT_DIR");
+  public static String drillReportDir = drillProperties.get("DRILL_REPORTS_DIR");
```
Agirish: That was a typo. I'll correct it.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -215,6 +212,14 @@ public void generateReports(List<DrillTest> tests, int iteration) {
       }
       bufferedWriter.flush();
       bufferedWriter.close();
+
+      // Upload report to DFS if the drillReportsDFSDir variable is set
+      if (!drillReportsDFSDir.isEmpty()){
```
cchang738: I had the same question before. If drillReportsDFSDir is defined, we will have two copies of the report files. My question is is it possible to have only one copy? Imagine I only want to getup my test environment to keep one copy of the report on DFS. Is it possible?

Agirish: This is done. The local report file is removed automatically after it's copied into DFS.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -175,33 +174,39 @@ public static void main(String[] args) throws Exception {
...
+
     try{
-      if(drillReportDirName == null){
-        drillReportDirName = CWD;
+      if(drillReportsDir == null){
+        drillReportsDir = CWD;
       }
-      File drillReportDir = new File(drillReportDirName);
+
+      File drillReportDir = new File(TestDriver.drillReportsDir);
```
cchang738: Is it really necessary to qualify drillReportsDir with TestDriver here?

Agirish: I'm not sure why I did that - you are right. Will update it.

Agirish: Done.

Agirish:
@cchang738 I'm going to have this committed if you don't have any more concerns.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -175,33 +174,39 @@ public static void main(String[] args) throws Exception {
...
               "report_" + new Date().toString().replace(' ', '_').replace(':','_') + ".json");
+
       BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(reportFile));
       Document document;
       for (DrillTest test : tests) {
         document = Json.newDocument();
         document.set("_id", test.getTestId()+ "_" + new File(test.getInputFile()).getName() + "_" + test.getCloneId() + "_" + iteration);
-        document.set("testId", test.getTestId());
-        document.set("queryFilepath", test.getInputFile());
+        document.set("queryFilepath", test.getInputFile().substring(test.getInputFile().indexOf("resources/")+10));
```
cchang738: I think you are trying to limit query file path length to 10. What's the reason for doing that?

Agirish: No - look carefully. What this is doing is to only consider the meaningful portion of the path - that's relative to the framework/resources directory. 

For example:
`/home/user1/drill-test-framework/framework/resources/Functional/suite/query.sql` is only saved as `Functional/suite/query.sql`. It makes it more readable.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
cchang738: LGTM. +1

PR #236, opened 2016-11-26
Updated the baselines to remove is.date.correct field from the metadata cache file

rchallapalli:
Abhishek & Robert,

Can one of you review this change?

Rahul


Agirish:
+1. Do we also need to eventually update the version from 1.9.0-SNAPSHOT to 1.9.0 for the 1.9.0 branch and 1.10.0-SNAPSHOT for master?

rchallapalli:
Yes... We should do that once the drill version changes.

Thanks for the review

PR #237, opened 2016-11-29
Fixing an NPE during test report generation

agirish:


framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -179,6 +179,9 @@ public void generateReports(List<DrillTest> tests, int iteration) {
       if(drillReportsDir == null){
         drillReportsDir = CWD;
       }
+      if(drillReportsDFSDir == null){
```
cchang738: Why check here, not at the place NPE is thrown? You just need to have one check right?

Agirish: Good point. I was trying to keep drillReportsDFSDir check consistent with drillReportsDir. Have updated it accordingly.

cchang738:
+1. thanks for the fix.

PR #239, opened 2016-12-01
Added tests for DRILL-4825 and DRILL-4823

rchallapalli:
Chun,

Below is the summary of this PR
1. Automated tests for Drill-4825 and Drill-4823
2. Changed the drill version number in the metadata cache expected values

Can you kindly take a look and let me know your thoughts?


framework/resources/Functional/metadata_caching/generated_caches/validate_cache1.e_tsv
```diff
@@ -1 +1 @@
...
...ionLevel":"0"\},"double_col":\{"name":\["double_col"\],"primitiveType":"DOUBLE","precision":"0","scale":"0","repetitionLevel":"0","definitionLevel":"0"\},"float_col":\{"name":\["float_col"\],"primitiveType":"FLOAT","precision":"0","scale":"0","repetitionLevel":"0","definitionLevel":"0"\}\}	\[\{"path":"/drill/testdata/metadata_caching/fewtypes/fewtypes.parquet","length":"2457","rowGroups":\[\{"start":"4","length":"1661","rowCount":"21","hostAffinity".*"columns":\[\{"name":\["int_col"\],"minValue":"1","maxValue":"21","nulls":"0"\},\{"name":\["bigint_col"\],"minValue":"302","maxValue":"98723980547","nulls":"0"\},\{"name":\["date_col"\],"minValue":"-4650","maxValue":"25667","nulls":"0"\},\{"name":\["time_col"\],"minValue":"0","maxValue":"86399990","nulls":"0"\},\{"name":\["timestamp_col"\],"minValue":"-2299559279000","maxValue":"4011874321000","nulls":"0"\},\{"name":\["interval_col"\]\},\{"name":\["varchar_col"\]\},\{"name":\["float_col"\],"minValue":"23.5","maxValue":"2.34555238E9","nulls":"0"\},\{"name":\["double_col"\],"minValue":"654.4","maxValue":"2.22223454324544E12","nulls":"0"\},\{"name":\["bool_col"\],"minValue":"false","maxValue":"true","nulls":"0"\}\]\}\]\}\]	1.10.0-SNAPSHOT
```
cchang738: Rahul, do you know if drillbit use this version number in any way? For example, do drillbit behave differently based on different version number? If answer is no, I don't see the need to keep verifying it.

cchang738:
Other than the comment I had, LGTM. +1

framework/resources/Functional/metadata_caching/generated_caches/validate_cache1.e_tsv
rchallapalli: If we write a wrong version number, then drill could potentially mis-interpret the dates in parquet. And also there is an email thread around this in the dev mailing list. Kindly take a look.

cchang738: The problem with this change is that test will fail if we test an older version of drill. We will have to strictly follow version control so framework version used is paired with drill version tested.

rchallapalli: We should be anyway pairing the framework version with the version of drill. Otherwise testing EBF's on older branches will be challenging.

MD-1181 would definitely help. But since this is a minor change which we have to make once in 3 months, I don't see it as super important right now.

Agirish: There are two reasons why MD-1181 is important. One is the effort needed every 3 months or so. But if you are willing to do that, then yeah, its okay to delay. 

But during a release of say 1.10.0, for a few weeks, we will have both 1.10.0-SNAPSHOT and 1.10.0 exist - one in the master and one in the release branch. This means during every release we will have these as false negatives. We will not be able to update baseline to 1.10.0 - and will need to wait until master moves to 1.11.0-SNAPSHOT before we can update baselines.

rchallapalli: Good point abhishek. You should add your comment to the jira directly. Also none of these comments should hinder the merge of this PR. Let me know if you think otherwise.

Agirish:
+1

PR #243, opened 2016-12-08
Test Framework changes for parquet filter pushdown project.

rhou1:
Hi Chun,

Can you review these changes?

Thanks.

--Robert

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
@@ -485,7 +485,9 @@ public class TestVerifier {
     String actual = new String(Files.readAllBytes(Paths.get(actualOutput)));
     boolean verified = false;
     if (verificationTypes.get(0).equalsIgnoreCase("regex")) {
-      verified = matchesAll(actual, expected);
+      verified = matchesAll(actual, expected, false);
+    } else if (verificationTypes.get(0).equalsIgnoreCase("regex-no-order")) {
+      verified = matchesAll(actual, expected, true);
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
```diff
@@ -518,6 +520,30 @@ private static boolean matchesAll(String actual, String expected) {
...
+   * @return true or false to indicate if there is a match
+   */
+  private static boolean matchesAllNoOrder(String actual, String expected) {
+    String[] expectedLines = expected.split("\n");
+    actual = actual.trim();
+    int i = 0;
+    for (String string : expectedLines) {
+      string = string.trim();
+      Matcher matcher = Pattern.compile(string).matcher(actual);
+      if (!matcher.find()) {
```
cchang738: Let's test with this code and see if it works. It may not though.

cchang738: Also, if this works, I recommend we just modify the existing matchesAll() to accept a "checkOrder" boolean param.

rhou1: I think this misses one case, where the result has more partitions than needed.  If the query should scan 3 partitions, but ends up scanning 4 partitions, I don't see a way to catch this with the new code.  So I think we need different methods depending on the use case.  I think it is hard to create one method that will work in all cases.

With the original code and the original use case (parquet filter pushdown), we check for extra partitions, missing partitions, and repeated partitions.

PR #244, opened 2016-12-13
Added tests to verify - CASE expression with constants

khurram-faraaz:
Hello Abhishek,

Can you please review and merge these new tests that verify case expressions with constants.
This is the first set of tests. I will add some more in a later commit.
I have a clean run, I will share the link to clean precommit run.

Thanks,
Khurram


Agirish:
LGTM. +1

kfaraaz:
Thanks Abhishek.

PR #249, opened 2017-01-16
Shell-scripts for dynamic UDFs testing.

KulykRoman:
All test cases which we can not run in test framework.


rchallapalli:
This is not the right repository to merge these changes as the scripts contain commands which are specific to mapr. Lets move these to a mapr specific repository. Few more general comments

1. Do not hard code the path "/user/root/**drill**/udf". On my cluster zk.root is not drill(its drill-build instead) and hence the scripts will fail. If this cannot be scripted, let me know
2. Modify the script to create "logs", "results" directories if they do not exist or update the readme file
3. When I ran the tests on a single cluster some tests failed, I will have to look into them

PR #250, opened 2017-01-17
Test Framework changes to enable some order-by clause verification to…

rhou1:
… work for JSON strings.

Chun,

Can you review the changes?

Thanks.

--Robert

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
@@ -30,6 +30,7 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.sql.Types;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -38,6 +39,7 @@ import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.Lists;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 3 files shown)


rhou1:
Simplified the code that parses the JSON string.

Added some more documentation on what queries can be verified by the code.

framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
```diff
@@ -348,16 +349,38 @@ private static int getMissingCount(Map<ColumnList, Integer> map) {
...
+   * order-by column, then the second order-by column is checked.  The second
+   * order-by column (and subsequent order-by columns) are handled in a recursive
+   * call to this procedure.
+   * Numerical values are handled separately from other values.
+   * JSON strings have special handling because they have to be parsed.  A field
+   * within a JSON string can be used in an order-by clause, so the field needs
+   * to be extracted from the JSON string.
+   * This code currently handles integer and string values in JSON strings.
+   */
+  private static int compareTo(ColumnList list1, ColumnList list2,
```
cchang738: @rhou1 This part of the code is never reached (code has been abandoned I believe). Let's seat down and take a closer look.

cchang738: I was mistaken on this part. Code is reachable.

framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
```diff
@@ -348,16 +349,38 @@ private static int getMissingCount(Map<ColumnList, Integer> map) {
   }
 
   private static int compareTo(ColumnList list1, ColumnList list2,
-      Map<Integer, String> columnIndexAndOrder, int start) {
```
rhou1: Added code to preserve original method signatures.

rhou1:
Here is the Jenkins run:

http://10.10.104.91:8080/job/Framework-Regression/894/consoleFull

framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
rhou1: The new method signatures use lists instead of maps to support duplicate column names when json string fields are used in the order-by clause.  The code that is added to preserve the original method signatures mostly convert the original maps into lists, and vice versa.

cchang738: This looks good now. @rhou1 can you take a look at the readme file and see anything there need to be updated after this change. Specifically, when we construct queries with order by clause, what are the rules we should follow? Would be nice if you can give some example query in the readme. Thanks.

rhou1: There are comments for verifyResultSetOrders that provide some of the rules.  Is this sufficient?  I can add some example queries.

rhou1:
Added documentation to README.md for validating SQL statements that have an order-by clause.

cchang738:
LGTM. +1. @rhou1  thanks for the effort.

rhou1:
@cchang738 Thanks for the review.

PR #254, opened 2017-01-23
Adding support for Simba JDBC Driver

rhou1:
Hi Abhishek,

Can you take a look at these changes?

I modified the json files to include a new category for simba jdbc tests.  I also added code to the test framework to handle json files that have not been updated to include the new category.  We can discuss which way is better, or have both options.

I will send the json file changes and test result changes in a separate pull request.

Thanks.

--Robert

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
@@ -30,10 +30,19 @@ import java.util.List;
 public class ColumnList {
   private final List<Object> values;
   private final List<Integer> types;
+  private final boolean Simba;
+  public static final String SIMBA_JDBC = "sjdbc";
 
   public ColumnList(List<Integer> types, List<Object> values) {
     this.values = values;
     this.types = types;
+    if (TestDriver.cmdParam.driverExt != null &&
+        TestDriver.cmdParam.driverExt.equals(ColumnList.SIMBA_JDBC)) {
+      this.Simba = true;
+    }
+    else {
+      this.Simba = false;
+    }
   }
 
   public List<Object> getValues() {
@@ -72,6 +81,7 @@ public class ColumnList {
       case Types.FLOAT:
       case Types.DOUBLE:
       case Types.DECIMAL:
+      case Types.REAL:
         break;
       default:
         hash += values.get(i).hashCode();
@@ -88,7 +98,35 @@ public class ColumnList {
   public String toString() {
     StringBuilder sb = new StringBuilder();
     for (int i = 0; i < values.size() - 1; i++) {
+      int type = (Integer) (types.get(i));
+      if (Simba && (type == Types.VARCHAR)) {
+        String s1 = String.valueOf(values.get(i));
+        // if the field has a JSON string or list, then remove newlines so that
+        // each record fits on a single line in the actual output file.
+        // sometimes Drill returns records in different orders, and by
+        // ensuring each reoord is on a single line, they can be matched in any
+        // order
+        if ((s1.length() > 0) &&
+           ((s1.charAt(0) == '{') || (s1.charAt(0) == '[')) ) {
+          s1 = Utils.removeNewLines(s1);
+          values.set(i, s1);
+        }
+      }
       sb.append(values.get(i) + "\t");
+    }
+    int type = (Integer) (types.get(values.size()-1));
+    if (Simba && (type == Types.VARCHAR)) {
+      String s1 = String.valueOf(values.get(values.size()-1));
+      // if the field has a JSON string or list, then remove newlines so that
+      // each record fits on a single line in the actual output file.
+      // sometimes Drill returns records in different orders, and by
+      // ensuring each reoord is on a single line, they can be matched in any
+      // order
+      if ((s1.length() > 0) &&
+         ((s1.charAt(0) == '{') || (s1.charAt(0) == '[')) ) {
+        s1 = Utils.removeNewLines(s1);
+        values.set(values.size()-1, s1);
+      }
     }
     sb.append(values.get(values.size() - 1));
     return sb.toString();
@@ -113,6 +151,7 @@ public class ColumnList {
       try {
         switch (type) {
         case Types.FLOAT:
+        case Types.REAL:
           float f1 = (Float) list1.get(i);
           float f2 = (Float) list2.get(i);
           if ((f1 + f2) / 2 != 0) {
--- a/framework/src/main/java/org/apache/drill/test/framework/ConnectionPool.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/ConnectionPool.java
@@ -19,6 +19,7 @@ package org.apache.drill.test.framework;
 
 import com.google.common.collect.Queues;
 //import org.apache.drill.jdbc.Driver;
+//import com.mapr.drill.jdbc41.Driver;
 import org.apache.log4j.Logger;
 import java.sql.Connection;
 import java.sql.DriverManager;
--- a/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
@@ -115,7 +115,6 @@ public class DrillTestJdbc implements DrillTest {
       query = queries[mainQueryIndex];
       executeQuery(query);
       
-
       if (getTestStatus()!= testStatus.CANCELED) { //Not to verify again if deliberately cancelled  
       	testVerifier = new TestVerifier(columnTypes, query, columnLabels, matrix.verificationTypes);
      	if (query.startsWith("explain") || matrix.verificationTypes.get(0).equalsIgnoreCase("regex") ||
@@ -123,6 +122,7 @@ public class DrillTestJdbc implements DrillTest {
           	matrix.verificationTypes.get(0).equalsIgnoreCase("filter-ratio")) {
         	setTestStatus(testVerifier.verifyTextPlan(modeler.expectedFilename, outputFilename));
       	} else {
+                // "in-memory"
         	setTestStatus(testVerifier.verifyResultSet(modeler.expectedFilename, outputFilename));
       	}
       }
--- a/framework/src/main/java/org/apache/drill/test/framework/Utils.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/Utils.java
@@ -39,6 +39,7 @@ import java.sql.Statement;
 import java.sql.Types;
 import java.sql.ResultSetMetaData;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -46,7 +47,6 @@ import java.util.PropertyResourceBundle;
 import java.util.ResourceBundle;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.Arrays;
 
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonNode;
@@ -228,10 +228,16 @@ public class Utils implements DrillDefaults {
 	    {
 	      failExtension = FilenameUtils.getExtension(failExtension);	
 	    }
-	    queryFileExtension = ".*."+failExtension;
+            // add queryFileExtension.substring to include ".q" or ".sql"
+            // this prevents expected results files with the "fail" extension from being picked up
+            // so queryFileExtension now looks like this: "*.q.fail"
+            queryFileExtension = ".*"+queryFileExtension.substring(2)+"."+failExtension;
 	  }
 	  else{ 
-	    queryFileExtension =    ".*.(fail|failing)";
+            // add queryFileExtension.substring to include ".q" or ".sql"
+            // this prevents expected results files with the "fail" extension from being picked up
+            // so queryFileExtension now looks like this: "*.q.fail"
+            queryFileExtension =    ".*" + queryFileExtension.substring(2) + ".(fail|failing)";
 	  }
         }
         boolean skipSuite = false;
@@ -243,14 +249,17 @@ public class Utils implements DrillDefaults {
           }
         }
         if (skipSuite) {continue;}
-          List<File> testQueryFiles = searchFiles(testDefFile.getParentFile(),
-                  queryFileExtension);
-          for (File testQueryFile : testQueryFiles) {
-	    //Expected File to find based on the original query Extension
-	    String expectedFileName = getExpectedFile(testQueryFile.getAbsolutePath(),
-                      originalQueryFileExtension, expectedFileExtension);
+
+        List<File> testQueryFiles = searchFiles(testDefFile.getParentFile(),
+                queryFileExtension);
+        for (File testQueryFile : testQueryFiles) {
+          //Expected File to find based on the original query Extension
+          String expectedFileName = getExpectedFile(testQueryFile.getAbsolutePath(),
+                    originalQueryFileExtension, expectedFileExtension);
+          if (expectedFileName != null) {
             drillTestCases.add(new DrillTestCase(modeler, testQueryFile.getAbsolutePath(), expectedFileName));
           }
+        }
       }
     }
     if (drillTestCases.size() == 0) {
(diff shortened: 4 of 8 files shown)


Agirish:
Thanks Robert. I'm taking a look now. 

@cchang738, can you also please review?

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -88,8 +93,36 @@ public int hashCode() {
...
+      if (categories.contains("smoke_simba_jdbc") && (type == Types.VARCHAR)) {
+        String s1 = String.valueOf(values.get(i));
+        // if the field has a JSON string or list, then remove newlines so that
+        // each record fits on a single line in the actual output file.
+        // sometimes Drill returns records in different orders, and by
+        // ensuring each reoord is on a single line, they can be matched in any
+        // order
+        if ((s1.length() > 0) &&
+           ((s1.charAt(0) == '{') || (s1.charAt(0) == '[')) ) {
+          s1 = s1.replace("\n","");
```
Agirish: Have you considered replacing "\n" (new line) with " " (space) instead of "" (empty string)? I'm concerned some strings with new lines will end up concatenated instead of simple eliminating new lines. 
For ex: 
"How are
you" 
should not end up as "How areyou", but instead as "How are you".

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -88,8 +93,36 @@ public int hashCode() {
...
+          values.set(i, s1);
+        }
+      }
       sb.append(values.get(i) + "\t");
     }
+    int type = (Integer) (types.get(values.size()-1));
+    if (categories.contains("smoke_simba_jdbc") && (type == Types.VARCHAR)) {
```
Agirish: Same as above - replace text with constant.

framework/src/main/java/org/apache/drill/test/framework/ConnectionPool.java
```diff
@@ -19,6 +19,7 @@
 
 import com.google.common.collect.Queues;
 //import org.apache.drill.jdbc.Driver;
+//import com.mapr.drill.jdbc41.Driver;
```
Agirish: Any reason for having commented out imports?

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -100,12 +100,15 @@ public void run() {
...
       executeQuery(query);
       
-      testVerifier = new TestVerifier(columnTypes, query, columnLabels, matrix.verificationTypes);
+      testVerifier = new TestVerifier(columnTypes, query, columnLabels, matrix.verificationTypes, modeler.categories);
       if (query.startsWith("explain") || matrix.verificationTypes.get(0).equalsIgnoreCase("regex") ||
           matrix.verificationTypes.get(0).equalsIgnoreCase("regex-no-order") ||
           matrix.verificationTypes.get(0).equalsIgnoreCase("filter-ratio")) {
         setTestStatus(testVerifier.verifyTextPlan(modeler.expectedFilename, outputFilename));
+      } else if (matrix.verificationTypes.get(0).equalsIgnoreCase("text") ) {
+        setTestStatus(testVerifier.verifyText(modeler.expectedFilename, outputFilename));
```
Agirish: Is "text" a new verification type? Can we have more comments on each of the verification types?

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -169,7 +170,21 @@ public static ConnectionPool getConnectionPool() {
...
+                testGroup.equals("smoke_simba_jdbc") &&
+                !testDefFile.getName().endsWith("apache.json") &&
+                !testDefFile.getName().endsWith("simba.jdbc.json")) {
+              // this json file applies to all Drivers.
+              // this json file is used for smoke tests
+              // this json file has not been modified to include
+              //    smoke_simba_jdbc tests
+              // therefore assume this json file should be modified to include
+              //    smoke_simba_jdbc tests
+              // therefore check for smoke_simba_jdbc tests
```
Agirish: Can we please clean-up the comments? And include clear details on each of the test definition file types (apache.json, simba.jdbc.json, etc.)? 

Also, may be consider renaming apache.json to either "apache.jdbc.json" or just the default ".json" to maintain backward compatibility.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -169,7 +170,21 @@ public static ConnectionPool getConnectionPool() {
         boolean foundTests = false;
         for (String testGroup : testGroups) {
           if (categories != null && !categories.contains(testGroup)) {
-            continue;
+            if (categories.contains("smoke") &&
```
Agirish: Let's not use "smoke" as the category, as it becomes restrictive and also "smoke" isn't well defined by us.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
Agirish: Also can you please add comments on backward compatibility? Are we saying all existing json files must be renamed to end with apache.json?

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -88,8 +93,36 @@ public int hashCode() {
   public String toString() {
     StringBuilder sb = new StringBuilder();
     for (int i = 0; i < values.size() - 1; i++) {
+      int type = (Integer) (types.get(i));
+      if (categories.contains("smoke_simba_jdbc") && (type == Types.VARCHAR)) {
```
rhou1: Sure.  I was thinking that the Simba tests would be parallel to the smoke tests.

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
rhou1: I don't think a JSON string can have a "\n" in it.  A JSON string can have "\\n" in it, but I believe JSON strings have to be "well-formed".

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
rhou1: Sure.

framework/src/main/java/org/apache/drill/test/framework/ConnectionPool.java
rhou1: The import is used for testing.  The line in question is for the Simba driver.  The line above it is for the Apache driver.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
rhou1: It is a new verification type.  I was originally going to use it for the JDBC tests, but I don't need it for that any more.  I am now planning to use it for orderby testing.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
rhou1: Perhaps we should address this too, then.  I went through almost all of the json files, and I saw mostly smoke.  I saw a few negative and other ones, but very few.  I would be happy to go through them all again and make them more consistent if possible.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
rhou1: I thought it would be more helpful if the comments could explain the code step-by-step.  But I can write a general comment.

I was wondering where to document the test definition file types.  Perhaps this is a good place for it.

Yes, I can rename apache.json to apache.jdbc.json.

Backward compatibilty is maintained.  By default, our current json files stay.  I have added two more categories (for jdbc and odbc) to each json file.  Most tests pass without any changes for Apache and Simba drivers.  A small number do not.  For these tests, I have renamed the query files to use a new file extension (*.q2) and created new json files for them.  There is a json file that Apache uses for these special tests, and a json file that Simba uses for these special tests.  Apache expected results files remain as they were.  I added new Simba expected results files with a new file extension (*.sjdbc.e_tsv as an example).

Having said that, I am concerned it is error-prone to require new json files to have additional categories for Simba.  So I added this code to enable existing json files to run with Simba without any changes.  When we had our last discussions, this was considered not the best approach.  So I have implemented both, so we can see what the impact is.

Perhaps I should add a few json files and test files so you can see how the whole thing works together.  Initially I just put the code here because it seems useful to have code separated from the rest of the changes, when the changes are large.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
rhou1: Counted 321 test definition files.

droptable suite uses droptable (2 json files)
maprdr suite uses functional_maprdb (1 json file)
some aggregate tests use "functional,smoke" (4 json files, including 2 under limit0)
impersonation suite uses impersonation (15 json files)
some suites use negative (10 json files)
complex uses regression (9 json files)
the rest use smoke (280 json files)

rhou1: The reason "smoke" is used as a category here is to show another way to incorporate the Simba drivers.  The one that we agreed upon is to add the Simba categories to the json files.  But this requires constant maintenance as we add more json files.  At the time, it was preferred not to make more changes to the framework than necessary.

So I decided to prototype what change would be needed to automatically invoke the Simba categories without changing the json files, to better evaluate the complexity.  In this case, I'm using "smoke" as a reference.  If a json file uses the "smoke" category, the framework will automatically add the Simba category.

We can go either way.  I have made all the changes to the json files, which will be in a separate PR.

rhou1:
Updated some code to address comments.

Added changes to table_function test suite as an example of how test definition files, query files and expected output files are changed.

framework/src/main/java/org/apache/drill/test/framework/ConnectionPool.java
Agirish: Okay thanks!

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
Agirish: Can we separate this out into a new PR?

framework/src/main/java/org/apache/drill/test/framework/Utils.java
Agirish: I think having a single way of defining "simba_jdbc" category is sufficient. Let's not use "smoke". Any test suite which needs to be added or updated for Simba JDBC should define the "simba_jdbc" category. I'm using it similarly for MapR-DB tests - they are categorized as "functional_maprdb".

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
Agirish: I'm still a little unclear on this. Have you considered JSON data with multi-line text? Are we replacing all newlines with ""?

framework/src/main/java/org/apache/drill/test/framework/Utils.java
rhou1: I think there is a difference between functional_maprdb and simba_jdbc.  Most tests should be run with simba_jdbc.  Most tests should not run with functional_maprdb.  I think the majority of the test suites should run with simba_jdbc.  So why do this manually?  Why not have the test framework handle this for us?

rhou1: We can have a group discussion about this when you get back.

rhou1: The main concern I have is, how do we know if we missed simba_jdbc somewhere.  I think it would be better to run with it "everywhere" and disable it when necessary.  Rather than add it when we remember to add it, which means we may miss it in some places.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
rhou1: Sure, I can pull the code out and put it in a new PR.  I just thought I would save some paperwork.

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
rhou1: I don't think you can have a new line break embedded within a string.

http://stackoverflow.com/questions/2392766/multiline-strings-in-json

Yes, I'm replacing all newlines with "".  I could replace them with " ", but I'm not sure how much of a difference it would make.

rhou1: Perhaps you can give me an example of what you are thinking about, in case I mis-understand your question.

rhou1:
Removed code for new verifyText method.  Will create a new PR so it can be used with orderby tests.

cchang738:
@rhou1 Let's discuss this PR in person. Thanks.

rhou1:
Had discussions with Chun.  Re-open PR for further discussions with the group.

rhou1:
Re-open pull request.

rhou1:
Force-push latest content to bring pull request back to the current state.

cchang738:
@rhou1 the proposed approach is rather messy. You need to move certain existing query files from .q to .q2; add a smoke-jdbc tag to certain .json files but not for some others; add another set of .json files to point to different baselines. Running tests with different drivers and maintaining tests can be confusing too.

Here is what I think we should do:

1. make no changes to query file names and test definition file (.json file)
2. add second sets of baseline with a predefined extension, such as *.e-mapr 
3. add a run time parameter such as -o to pick up an alternative baseline file if exists based on extension (such as *.e-mapr and can be specified at commandline).

This is hopefully a much simpler approach. When we need to test an alternative driver, just give the -o option and the framework will automatically pick the right baseline.

rhou1:
@cchang738 
I used the approach in this PR to maintain the flexibility provided by the framework.  And this PR provides a way to avoid adding the smoke_simba_jdbc tag to certain json files (by running smoke_simba_jdbc tests wherever smoke tests are run).  Adding the smoke_simba_jdbc tag was the original design that we agreed to, but I agree it can be messy.  Which is why I added a way to avoid it.

Having said that, I think your approach can work too, although it is not as flexible in some ways.   And if we need the extra flexibility, we could use the approach in this PR for those specific cases.  So I'm happy to go with either approach.

rhou1:
We also need to find a way to accommodate tests that cannot be run with Simba.  They need to be run with Apache, and disabled to run with Simba.  We might be able to use another file extension to identify these tests.

rhou1:
Cleaned up the code based on group discussion some time ago.  Abhishek, Chun, please take a look.

rhou1:
Added comments to clarify why queryFileExtension.substring(2).  Before this change, the queryFileExtention was "*.fail".  Now the queryFileExtension is "*.q.fail".  This enables the ".fail" extension to be used with other files, such as expected results files.  When an expected results file has a ".fail" extension, such as "*.e_tsv.sjdbc.fail", the corresponding test will not be executed with the Simba JDBC driver due to an existing bug.

rhou1:
Added removeNewLines function to remove new line characters (\n) except when they exist in a string demarcated by double quotes.

This allows us to handle a case like this:
0: jdbc:drill:drillbit=10.10.100.186> select * from `newline5.csv`;
+--------------------------------+
|            columns             |
+--------------------------------+
| ["line1"]                      |
| ["{line2\nabc\ndef\n}"]        |
| ["line3"]                      |
| ["{line4\n   abc\n   def\n}"]  |
+--------------------------------+

This is a parquet file with strings.  Here, the string does not have double quotes.  The code cannot handle this case, but this is an issue for the test framework in general, because we need each record to be on a single line.
0: jdbc:drill:drillbit=10.10.100.186> select * from `newline7`;
+-------------------------+
|         EXPR$0          |
+-------------------------+
| line1                   |
| {line2
abc
def
}        |
| line3                   |
| {line4
   abc
   def
}  |
+-------------------------+

cchang738:
+1 LGTM. But please run pre-commit test with latest change if you haven't done so.

Agirish:
+1. Let's get this in, as it's been pending from long. We can always work on improving anything on a later date.

rhou1:
I re-ran the pre-commit tests with the latest changes.

rhou1:
Chun,  Abhishek, thanks for the review.

PR #256, opened 2017-01-31
Tests for parquet filter pushdown project.

rhou1:
Hi Rahul,

Can you review these?

Thanks.

--Robert


rchallapalli:
Review still in progress....

At a glance, the tests look good. I am curious to know how you generated the baselines. If you manually verified the baselines, then I would express some concern and would request you to use mysql or postgres to generate them.

rhou1:
I manually verified the baselines.   The tests were designed to generate certain answers so that I could verify them.  So for example, I created a test that would return 3020 rows, and it returned 3020 rows.

PR #258, opened 2017-02-06
DRILL-5237: add test coverage

cchang738:
Add test coverage for a case when flatten operator returns empty batch for the first time.

@rchallapalli Can you please review?

http://10.10.104.91:8080/job/Framework-Regression/896/console


rchallapalli:
I see that the query passed in your functional run but is the below even syntactically correct?

"where t1.a .a1 like 'a1'"

cchang738:
@rchallapalli good catch. It actually works as is but I made the change. Below is the test run after the change.

http://10.10.104.91:8080/job/Framework-Regression/897/console

rchallapalli:
LGTM +1

PR #259, opened 2017-02-10
MD-1181: Support injecting variables in regex baseline

cchang738:
Currently only supports {DRILL_VERSION}.

@rchallapalli Can you please review? Thanks.


rchallapalli:
Hacky....and if we want to inject more variables it becomes more hacky. I would suggest that you use an enum defining all variables that you want to inject and for all variables in the enum you can fetch their values from "drillProperties" or something around those lines. And then replace them in the expected file's string value

Also I would suggest that you modify at-least one existing test to demonstrate that the feature is working, otherwise the reviewer has to do that himself. You can choose any one test from the below suite

https://github.com/cchang738/drill-test-framework/tree/2e37257e4f79abc27f59bf918d9d208d7e6af684/framework/resources/Functional/metadata_caching/generated_caches

cchang738:
@rchallapalli Addressed your comments. 1. made it more extensible; 2. changed all baselines to use the injection parameter (verify if i missed any).

http://10.10.104.91:8080/job/Framework-Regression/901/console

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -65,7 +67,9 @@
   private Connection connection = null;
   private ConnectionPool connectionPool = null;
   private String commitId;
-  private String version;
+  private static String version;
+  private String[] injectionKeys = {"DRILL_VERSION"};
```
rchallapalli: I guess we should be making this variable final

rchallapalli: This is OK. However I think an enum is more appropriate for injectionKeys as we can avoid using the hardcoded constants ("DRILL_VERSION") directly in other places of the code

rchallapalli:
Thanks for addressing the comments chun and adding this feature. Also I appreciate you changing the baselines as well. 

Can you look at the minor comments I have? Apart from that it looks good overall

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
cchang738: Will change version to static final. enum type is not suited here. It will work well with the switch statement but not work well with String.replaceAll(). In addition, the injections HashMap needs String type as key.

cchang738:
@rchallapalli It turned out I can't use final here. You will get compilation errors. And as I mentioned, enum wouldn't work well here either. The injections map uses string as key. And String.replaceAll() uses String.

I did clean up the code a bit. Here is another run.

http://10.10.104.91:8080/job/Framework-Regression/903/console

Thanks for the review.

rchallapalli:
LGTM +1

Thanks for taking a look

PR #261, opened 2017-02-20
MD-1283: Show profile IDs for all the tests in the console output

kkhatua:
Added support to publish queryID for each executed query. This is published in addition to the connection ID.


cchang738:
@kkhatua Let's not creating branches directly on master repo. We clone and branch on our individual repo, making changes there and creating PR against individual branch.

kkhatua:
Ok. Do you want me to clone and create a PR from that branch? The patch was minor, so I just created a branch here itself.

Agirish:
That would be consistent with how we make fixes.

PR #264, opened 2017-03-16
MD-1497: fix copy source data dir

cchang738:
@Agirish please review this pull request. There is a typo in my last commit related to source data dir.


Agirish:
The variables DRILL_TESTDATA and DRILL_TEST_DATA_DIR mean different things, but since they are badly named, it's very confusing. We should consider renaming these to be more meaningful. Perhaps something like TESTCASE_DIR and DFS_TEST_DATA_DIR

+1 otherwise.

cchang738:
Agree with the names not clear. Since they are historical, I will merge this PR and make changes in another PR given time for thorough testing. Don't want to rush the changes since they may affect many places.

PR #272, opened 2017-04-14
Fix tests so that they do not update the metadata cache concurrently.…

rchallapalli:
Krystal,

I made sure that metadata cache tests do not fail randomly due to cache corruption. For this I have to 

1. Move "refresh table metadata" commands to datagen so that they do not run randomly
2. Disable tests which cause automatic cache refresh
3. Disable tests where the plan might change due to a change in order of reading files

 Rahul


krystaln:
LGTM +1

PR #273, opened 2017-04-14
Update README.md

agirish:
Remove the note on Regex Ordering. We can add this back when the feature is in. Also, the note needs to be better formatted, clear and concise.


cchang738:
I think this feature is in. The readme portion might be too verbose. Please check with @rhou1

Agirish:
I'll work on a design to have multiple README files, after which we can move this out, and link it to the main README. 

Reference: https://github.com/mapr/drill-test-framework/blob/master/framework/resources/Advanced/README.md

PR #275, opened 2017-04-20
Added tpcds tests against hive2.1 generated parquet files

rchallapalli:
Abhishek,

I added tpch queries to run on top of hive 2.1 generated parquet files. Added tests for hiveplugin, parquet reader, and hive plugin using the native parquet reader. Can you kindly take a look?

Rahul


framework/pom.xml
```diff
@@ -58,6 +58,7 @@
                   <outputFileName>tpcds_hivegenerated_parquet.tgz</outputFileName>
                   <unpack>true</unpack>
                   <outputDirectory>${project.basedir}/resources/Datasources</outputDirectory>
+                  <md5>b585a419fb5ceda6daa91a8bc19c2365</md5>
```
Agirish: Hey Rahul, is this to detect corrupt downloads?

rchallapalli: It serves 2 purposes

1. Detect corrupt downloads. So we might as well add to all other files as well
2. I updated the tgz file on amazon. However the maven plugin did not use it and used one from its cache. So having this md5 forced maven to download the latest

Agirish: That's great. We should definitely add this to all other files.

PR #279, opened 2017-04-21
Fix tests so that they do not update the metadata cache concurrently.…

rchallapalli:
Krystal,

I backported the metadata cache related changes to the 1.9.0 branch as well. Got a functional run with 2 known verification failures. Can you take a look?

Rahul


krystaln:
LGTM +1

PR #282, opened 2017-06-07
MD-1954 - Add support to run tests tagged as failures

kedarbcs16:
We have been following the convention to mark a test as failed by suffixing a ".fail" extension. These failed tests can be for features not checked-in, known bugs or similar.
This change adds a command line option to trigger running these failed tests and an option to remove the .fail extension if the test passes.


rhou1:
Has this code been tested?

rhou1:
Hey, this is a great start!  Thanks for doing this.

I would test it with a few different test cases.

I would also consider enforcing a file naming convention.  Maybe rename all failed tests so that they end in ".fail" or something.  This is something that would be done manually using git.  It will make your code a little simpler.  And it would make it easier to verify your testing.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -170,6 +170,15 @@
         
         String queryFileExtension = modeler.matrices.get(0).inputFile;
         String expectedFileExtension = modeler.matrices.get(0).expectedFile;
+        String tempQueryExt = queryFileExtension;
+        if(TestDriver.cmdParam.negativeCasesInclusion == true){
+            LOG.info(LINE_BREAK);
+            LOG.info("fail added");
+            LOG.info(LINE_BREAK);
+            queryFileExtension += "|"+tempQueryExt+".fail"+"|"+tempQueryExt+".failing";
```
Agirish: Use static final variables for literals. We could also look this up in the properties file.

rhou1:
Also, add comments to your PR as you work.

Agirish:
I agree with Robert. Lets assume every failed test as a .fail extension. Please test your code and show us some results (by sharing pre-commit test links over email).

Agirish:
Kedar, please rebase this branch and squash your commits (related to this fix).

PR #285, opened 2017-06-16
MD-2094: reset system options to default inside before-run.sql

cchang738:
Before automation run, we need to reset system options to default to ensure clean environment.

@rchallapalli Can you please review?


rchallapalli:
LGTM +1

PR #286, opened 2017-06-27
Removed the extra project operator from the baselines after the fix for drill-5538

rchallapalli:
Abhishek,

The fix for drill-5538 removes the un-necessary project from the plan. I made the necessary changes to the baselines. Can you please take a look?

Rahul


Agirish:
+1. LGTM

rchallapalli:
Thanks Abhishek!

PR #295, opened 2017-07-14
Clean-up: Rename 'smoke', 'regression' test group categories to 'functional'

kedarbcs16:


Agirish:
+1. Thanks for making this change.

PR #296, opened 2017-07-17
Removed fail extensions from tests that are passing now

kedarbcs16:
There are 2 commits in this :- 
1st - to change remove .fail from test cases that are marked as failure that are passing now
2nd - handle cases that are not having query-extensions in them (.q or .sql) and include them in the names


Agirish:
The file ending .q.drill4137 should be .q (or .sql - verify extension in test def file) instead.

Agirish:
Same for coalesce.q.drill-3862

kedarbcs16:
@Agirish  addressed comments ...

Agirish:
I'm assuming these tests are passing now :) Have you run any regression tests to confirm?

kedarbcs16:
yes these tests always pass ,I have run this only nodes several times

rhou1:
Can I see the Jenkins job that shows which tests now pass, and which ones are still failing?

Have you run the full Jenkins job?

kedarbcs16:
No have not run a jenkins job for a while now

Agirish:
+1. Let's get this in now. From what I discussed and see, this should be a safe check-in. And later make any changes if required.

PR #298, opened 2017-07-22
Increase planner.memory.max_query_memory_per_node

agirish:
The current default of 2 GB is now insufficient, as the 2 GB limit is being used by both Hash Agg and Managed Sort. As a result multiple queries fail with OOM. 

Temporarily changing this limit to a much higher value (10 GB), until the default is updated suitably (tracked by DRILL-3549).


arina-ielchiieva:
+1, LGTM.

Agirish:
Thanks @arina-ielchiieva 
@cchang738 , I've merged this in, but do take a look and let me know your comments - if any.

PR #300, opened 2017-08-06
Hive datagen

prasadns14:
Remove data generation dependency in hive test cases

### DIFF
--- a/framework/resources/Functional/hive/hive_functions/hive_functions.json
+++ b/framework/resources/Functional/hive/hive_functions/hive_functions.json
@@ -45,27 +45,27 @@
         },
         {
            "mode": "gen",
-           "src": "framework/resources/Datasources/hive_storage/execHive.sh framework/resources/Datasources/hive_storage/hive_storage.ddl",
+           "src": "Datasources/hive_storage/execHive.sh framework/resources/Datasources/hive_storage/hive_storage.ddl",
            "dest": ""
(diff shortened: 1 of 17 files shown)


Agirish:
Hey Prasad, thanks for making these changes. They should be very helpful. 

Can you please create a wrapper script for each test suite and have it call all relevant scripts? The reason being that if any of the test suites require the exact order of scripts to execute, then there may be issues. The wrapper script can be in corresponding Datasouces/<testSuite>.

Agirish:
Come to think of it, if multiple suites call the same set of scripts, then the wrapper script may be an issue. We'll need to make sure to none of the scripts have any cross dependencies - and not dependent on a prior script to have executed in advance. As during concurrent executions, ordering cannot be guaranteed.

prasadns14:
@Agirish - I ran the Functional, Functional/hbase and Functional/hive suites in separate runs(126 - 128). All the cases except a couple of hive(baseline issue) are passing.
There are no dependencies across suites. We can merge this pull request for now, and come up with a plan later.

cchang738:
I discussed this with @prasadns14. The proposed solution seems well thought out and works well. Let's have this merged. We can discuss and address other concerns in a separate PR if needed. Thanks.

LGTM. +1

Agirish:
Sure. +1. Thanks Prasad.

PR #303, opened 2017-08-14
MD-156: Support a marker ("@") in front of a query to mark the query …

cchang738:
…to be verified.

@kkhatua This is the fix to the current automation framework requirement that queries in a testcase must be odd number of lines. I added the option to have "@" as a marker for the query to be verified. And this change is backward compatible. Can you review this PR? I tested locally as well as running pre-commit test. Here is the link to pre-commit test run. Thanks.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
@@ -79,7 +79,7 @@ public class DrillTestJdbc implements DrillTest {
     final Stopwatch stopwatch = Stopwatch.createStarted();
     this.thread = Thread.currentThread();
     setTestStatus(TestStatus.RUNNING);
-    int mainQueryIndex = 0;
+    int mainQueryIndex = -1;
     String[] queries = null;
     try {
       connection = connectionPool.getOrCreateConnection(modeler);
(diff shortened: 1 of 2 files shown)


amansinha100:
U ggg uii Hz Hzh huh ink in

Get Outlook for Android<https://aka.ms/ghei36>

XXXC CCXXXXXTXTGXXXXCCXXXXXTXTGXXXXXC XXTXXXTXTXTXXTXTXTTTG GGTG

________________________________

amansinha100:
Tv tx f2fy gxxtxtftttgxxtxtftttff

Get Outlook for Android<https://aka.ms/ghei36>


________________________________

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -95,8 +95,16 @@ public void run() {
       }
 
       queries = Utils.getSqlStatements(modeler.queryFilename);
-      mainQueryIndex = queries.length / 2; // Currently, the main query must be in the middle of the list of queries
-
+      for (int i = 0; i < queries.length && mainQueryIndex == 0; i++) {
+    	if (queries[i].startsWith("@")) {
```
kkhatua: It might be worthwhile to check for `--@` instead of just `@`
This allows the file to be consumed by other SQL Engines too, since a leading double hyphen is treated by most SQL parsers as comments. 
You could then expand the annotations to do more advanced things, on similar lines of how test frameworks like JUnit, etc. interpret java code as elements of test - e.g. setup, test, cleanup.

Agirish: I agree. Let's make it --@

PR #306, opened 2017-08-23
Disabled queries related to hive datetime functions

prasadns14:
The result of hive datetime functions are dependent on the machine timezone. (DRILL-5002)

These cases are failing whenever drill runs in a non-PST timezone. I discussed the issue with @cchang738 and @krystaln, and it was decided to mark the queries as fail till the issue is resolved.


Agirish:
+1. Thanks Prasad.

PR #307, opened 2017-08-24
Marking ORC tests as failed

agirish:
We frequently observe that these tests fail. Marking them with .fail extension until we find a resolution. 

It's probably that the data generation script, run as a part of test preparation during a regression run, has an issue. Occasionally, the Hive job to load a Hive table -
which spawns a MapReduce job, ends up writing two output files, instead of overwriting/replacing the data file. This ends up in the table having twice as many rows as intended. We don't know what causes this native Hive job to malfunction - probably an issue with a race condition in Hive MR. Further investigation is needed.


krystaln:
LGTM +1.

PR #310, opened 2017-09-01
Add option to avoid running failed tests a second time.

rhou1:
The test framework runs failed tests a second time to see if they are random failures.  This option avoids running tests a second time.  It is helpful when there are a lot of failed tests, or if the failed tests take a long time to run.

Abhishek,  Chun, please take a look.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/CmdParam.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/CmdParam.java
@@ -55,6 +55,9 @@ import com.beust.jcommander.Parameter;
     @Parameter(names = {"-rf", "--runFailed"}, description = "Run tests tagged as failures", required=false)
     public boolean runFailed  = false;
 
+    @Parameter(names = {"-sR", "--skipRandom"}, description = "Skip check for random failures", required=false)
+    public boolean skipRandom  = false;
+
     public List<String> excludeDependenciesAsList() {
       if (excludeDependencies == null) {
         return new ArrayList<String>();
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -197,7 +197,7 @@ public class TestDriver implements DrillDefaults {
       
       for (DrillTest test : tests) {
         TestStatus testStatus = test.getTestStatus();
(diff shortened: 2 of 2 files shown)


cchang738:
I prefer by default, we do not automatically rerun failed test. I suggest

cchang738:
This is a feature I want. Thanks for doing it.

framework/src/main/java/org/apache/drill/test/framework/CmdParam.java
```diff
@@ -55,6 +55,9 @@
     @Parameter(names = {"-rf", "--runFailed"}, description = "Run tests tagged as failures", required=false)
     public boolean runFailed  = false;
 
+    @Parameter(names = {"-sR", "--skipRandom"}, description = "Skip check for random failures", required=false)
+    public boolean skipRandom  = false;
```
Agirish: Let's set the default to true.

Agirish: So it does not skip by default, correct? I got confused the first time i saw this. Only if the option is provided, then we skip.

framework/src/main/java/org/apache/drill/test/framework/CmdParam.java
```diff
@@ -55,6 +55,9 @@
     @Parameter(names = {"-rf", "--runFailed"}, description = "Run tests tagged as failures", required=false)
     public boolean runFailed  = false;
 
+    @Parameter(names = {"-sR", "--skipRandom"}, description = "Skip check for random failures", required=false)
```
Agirish: We need to fix the parameter names. We have too many and there is no clear convention being followed. Let's get this in for now, as is. Later work towards cleaning up.

Agirish:
+1. Thanks Robert.

framework/src/main/java/org/apache/drill/test/framework/CmdParam.java
rhou1: Correct.  If the option is provided, then we skip.

framework/src/main/java/org/apache/drill/test/framework/CmdParam.java
rhou1: Sure.

rhou1:
Thanks for the reviews and comments.

PR #313, opened 2017-09-06
Baseline changes related to DRILL-5747

prasadns14:
As part of the fix for DRILL-5747, there have been changes in the ordering of columns when * operator is used. I have fixed the baseline accordingly. The tests ran fine.

Please review.


cchang738:
Talked to @prasadns14, he used a script to fix baseline and pre-commit test past. Although not accessible out side MapR, I prefer we post pre-commit test link in our PR so it's documented and easily verified by MapR people. Thanks for the fix.

+1

Agirish:
I disagree. This is a public repository and it's not good to share private links. Please use emails to share the links. It's upto the reviewer and author to make sure appropriate tests were run. We do not need to monitor the same.

cchang738:
I disagree. This is a MapR repository made accessible to public. Think about it, it's not apache. And we should choose the most productive way to conduct our work.

PR #315, opened 2017-09-06
Separate data verification failures from plan verification failures.

rhou1:
Also add query ID to the status of each test.  The query ID can be used to find
the profile.  Each node has to be checked for the profile.  Here is a sample:

PASS (1.988 s) /root/drill-test-framework/framework/resources/Functional/window_functions/lag_func/lag_Fn_93.q (connection: 2007519911) (queryID: 26504575-863a-85bc-85d4-ae7a0ef3a8b9)

Execution failures may not have a query ID because the query did not complete.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
@@ -22,6 +22,7 @@ import com.google.common.collect.Lists;
 import org.apache.drill.test.framework.TestCaseModeler.TestMatrix;
 import org.apache.drill.test.framework.TestVerifier.TestStatus;
 import org.apache.drill.test.framework.TestVerifier.VerificationException;
+import org.apache.drill.test.framework.TestVerifier.PlanVerificationException;
 import org.apache.drill.test.framework.DrillDefaults;
 import org.apache.log4j.Logger;
 
@@ -63,6 +64,7 @@ public class DrillTestJdbc implements DrillTest {
   private AtomicBoolean doneProcessingResultSet = new AtomicBoolean(false);
   private int id;
   private int totalCases;
+  private String queryID;
 
   private static volatile int noOfCasesCompleted;
 
@@ -128,7 +130,9 @@ public class DrillTestJdbc implements DrillTest {
     	  executeLimitZeroQuery(limitZeroQuery);
       }
     } catch (VerificationException e) {
-      fail(TestStatus.VERIFICATION_FAILURE, e);
+      fail(TestStatus.DATA_VERIFICATION_FAILURE, e);
+    } catch (PlanVerificationException e) {
+      fail(TestStatus.PLAN_VERIFICATION_FAILURE, e);
     } catch (Exception e) {
       fail(TestStatus.EXECUTION_FAILURE, e);
 	} finally {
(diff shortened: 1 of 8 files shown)


cchang738:
Can you test with MapR JDBC driver and see if your change works? Thanks.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.drill.test.framework.TestCaseModeler.TestMatrix;
 import org.apache.drill.test.framework.TestVerifier.TestStatus;
 import org.apache.drill.test.framework.TestVerifier.VerificationException;
+import org.apache.drill.test.framework.TestVerifier.PlanVerificationException;
```
Agirish: Can we change the other to DataVerificationException?

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -162,6 +162,7 @@ public int runTests() throws Exception {
     int totalPassingTests = 0;
     int totalExecutionFailures = 0;
     int totalVerificationFailures = 0;
```
Agirish: May be change to totalDataVerificationFailures?

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -317,9 +337,11 @@ public int runTests() throws Exception {
       }
       LOG.info(LINE_BREAK);
 
-      LOG.info(String.format("\nPassing tests: %d\nExecution Failures: %d\nVerificationFailures: %d" +
+      LOG.info(String.format("\nPassing tests: %d\nExecution Failures: %d\nVerification Failures: %d" +
+        "\nPlan Verification Failures: %d" +
```
Agirish: May be change Verification Failures to Data Verification Failures?

Agirish:
+1. Thanks Robert. Have a few minor review comments.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
rhou1: I kept VerificationException because this is "open-source" and I was wondering if someone might expect "Verification Failure" to still work.

rhou1:
The time to run the tests is about the same.  The run was actually a minute faster than a previous run last week.

I ran with -rf.  I also created my own tests where I forced an Execution Failure, a Data Verification Failure and a Plan Verification Failure.  And some combinations of these three.

We don't support Simba JDBC driver at this time.  When I get to that, I'll make adjustments then.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
Agirish: I don't think that should impact any consumer - if they rebase, they should get the latest changes. Previous forks should condinue to work. We could always simply keep the VerificationException type, but not use it anywhere in the code.

cchang738:
We do support MapR JDBC driver. We need to make sure the proposed change does not break that.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
rhou1: Done.

rhou1:
Modified queryID code to use fully qualified class name.  Code needs to be compiled with Apache JDBC driver.  It can then be used with either Apache or Simba JDBC driver.

Agirish:
Thanks Robert. If there are no other concerns, we can have this merged.

cchang738:
LGTM +1

PR #318, opened 2017-09-09
MD-884: Changed plugin storage name from dfs to dfs_test

prasadns14:
@cchang738 and @Agirish Please review the changes


Agirish:
Thanks Prasad. 

You will also need to change the name of the file -> conf/plugin-templates/dfs-storage-plugin.template to conf/plugin-templates/dfs_test-storage-plugin.template

prasadns14:
@Agirish 
The cttas cases create temporary tables. Currently, temporary tables can be created only in the default temporary workspace dfs.tmp (https://gist.github.com/arina-ielchiieva/50158175867a18eee964b5ba36455fbf).

Should i revert the  workspace name for cttas cases to dfs.tmp? Or do you have other suggestions?

Agirish:
The name of the storage plugin shouldn't matter. dfs_test.tmp should be equivalent to dfs.tmp. Do you see any issues?

Agirish:
Also, I took a look at as many files I could - looks good. But with lot of files, Guthub makes it hard. Running these tests (alongside the corresponding mapr private tests) on the Docker framework should give us more confidence. There may be areas which we have missed updating / reviewing.

PR #324, opened 2017-09-19
MD-2296: New cases to run MapRDB binary cases with hbase storage plugin

prasadns14:
Duplicated the maprdb binary cases to run with hbase storage plugin.
@cchang738, please review


cchang738:
Please make sure "testId" specified in your .json file is unique and not the same as with the binary version. +1 after you make that change.

prasadns14:
@cchang738, i have changed the testId

PR #326, opened 2017-09-26
DRILL-4783: Fixed baseline related to flatten

prasadns14:
Fixed the baseline for flatten cases as per DRILL-4783 behavior.
@cchang738  please review


cchang738:
LGTM. +1

PR #328, opened 2017-10-03
Add option to enable test framework to run tests in a specified order.

rhou1:
The option is enabled with the -rR flag, for repeatRun.
If specified for the first time, it will create a file in framework/resources/queryLog.log
that contains the tests that are run, in the order they are run.
Once this file is created, the next run will run the tests in the order specified
in the file.  The file can also be created manually.
This implementation requires that the tests to be run are also specified by the -s option.
It takes the tests that are normally collected, and re-orders the tests in the order
specified in the queryLog.log file.
This is useful for failures that occur when tests are run in a particular order.
It is mainly useful if the tests are run without concurrency, meaning one at a time.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -36,8 +36,10 @@ import org.apache.log4j.Logger;
 import org.ojai.Document;
 import org.ojai.json.Json;
 
+import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.util.*;
@@ -189,6 +191,56 @@ public class TestDriver implements DrillDefaults {
 
       LOG.info("> RUNNING TESTS (ITERATION " + i + ")..");
       Collections.shuffle(tests, new Random());
+      List<DrillTest> new_tests = Lists.newArrayList();
+      String queryLogFilename = drillOutputDir + "/queryLog.log";
+      List<String> skippedTests = Lists.newArrayList();
+      File queryFile = new File(queryLogFilename);
+      if (cmdParam.repeatRun == true) {
+        // check if file has list of tests
+        if (queryFile.length() > 0) {
+          // generate a new list of tests in the requested order
+          try {
+            BufferedReader reader = new BufferedReader(new FileReader(queryFile));
+            boolean found = false;
+            String line = "";
+            while ((line = reader.readLine()) != null) {
+              for (DrillTest test : tests) {
+                // check that requested test is in the list of tests
+                if (found = test.getInputFile().equals(line)) {
+                  new_tests.add(test);
+                  break;
+                }
+              }
+              if (!found) {
+	        skippedTests.add(line);
+              }
+              found = false;
+            }
+            reader.close();
+            tests = new_tests;
+          } catch (Exception e) {
+            LOG.error("\nCould not open queryFile " + queryLogFilename + " containing tests to run");
+            LOG.error(e.getMessage());
+            System.exit(-1);
+          }
+        } else {
+          // file with list of tests does not exist.  create new file and
+          // record the tests and the order in which they are run
+          try {
+            BufferedWriter writer = new BufferedWriter(new FileWriter(queryFile));
+            if (writer != null) {
+              for (DrillTest test : tests) {
+                writer.write (test.getInputFile() + "\n");
+              }
+              writer.close();
+            }
+          } catch (Exception e) {
+            LOG.error("\nCould not create queryFile " + queryLogFilename + " to record which tests are run");
+            LOG.error(e.getMessage());
+            System.exit(-1);
+          }
+        }
+      }
       executor.executeAll(tests);
 
       if (cmdParam.trackMemory) {
@@ -277,14 +329,25 @@ public class TestDriver implements DrillDefaults {
         LOG.info(test.getInputFile());
         LOG.info("Query: \n" + test.getQuery());
       }
-      LOG.info("Random Failures:");
+      if(randomFailures.size()>0){
+        LOG.info("Random Failures:");
+      }
       for (DrillTest test : randomFailures) {
         LOG.info(test.getInputFile());
         LOG.info("Query: \n" + test.getQuery());
         LOG.info("Failed with exception", test.getException());
       }
+      if(skippedTests.size() > 0) {
+        LOG.info("Skipped Tests: These tests cannot be run from " + queryLogFilename + " because they were not in source or group, or it is a duplicate test");
+        for (String line : skippedTests) {
+          LOG.info(line);
+        }
+      }
       LOG.info(LINE_BREAK);
       LOG.info("Summary");
+      if ((cmdParam.repeatRun == true) && (queryFile.length() > 0)) {
+        LOG.info("Executed queries from " + queryFile);
+      }
       if(cmdParam.runFailed == true){
         if(passingTests.size()>0){
           LOG.info(LINE_BREAK);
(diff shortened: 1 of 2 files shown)


rhou1:
Chun, can you please review this.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -188,6 +190,38 @@ public int runTests() throws Exception {
...
+                    tests.add(test);
+                  }
+                  break;
+                }
+              }
+            }
+            reader.close();
+          }
+        }
+        BufferedWriter writer = new BufferedWriter(new FileWriter(new File(
```
cchang738: This can be optimized, i.e. we probably don't need to write to the file again if it exists already.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -188,6 +190,38 @@ public int runTests() throws Exception {
...
+                                queryLogFilename)));
+        if (writer != null) {
+          for (DrillTest test : tests) {
+            writer.write (test.getInputFile() + "\n");
+          }
+          writer.close();
+        }
+      }
```
cchang738: Might be better rearrange the flow a bit as we discussed in person. Thanks.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -188,6 +190,38 @@ public int runTests() throws Exception {
 
       LOG.info("> RUNNING TESTS (ITERATION " + i + ")..");
       Collections.shuffle(tests, new Random());
+      if (cmdParam.repeatRun == true) {
+        String queryLogFilename = drillTestDataDir + "/queryLog.log";
```
cchang738: Can we store the query file inside drill-out dir? This is where we put other tmp files, so keep it consistent would be better.

rhou1:
Hi  Chun,

I have addressed your comments.  Please take a look and let me know if I missed something.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -187,7 +189,47 @@ public int runTests() throws Exception {
       stopwatch.reset().start();
 
       LOG.info("> RUNNING TESTS (ITERATION " + i + ")..");
-      Collections.shuffle(tests, new Random());
+      String queryLogFilename = drillOutputDir + "/queryLog.log";
+      File file = new File(queryLogFilename);
+      if ((cmdParam.repeatRun == true) && (file.exists() && (file.length() > 0))) {
+        // run the tests in the requested order
+        try (BufferedReader reader = new BufferedReader(new FileReader(new File(
```
cchang738: Can we reuse the "file" already created two lines above?

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -187,7 +189,47 @@ public int runTests() throws Exception {
...
+          LOG.error("\nCould not open file containing tests to run");
+          LOG.error(e.getMessage());
+          System.exit(-1);
+        }
+      } else {
+        // shuffle the tests
+        Collections.shuffle(tests, new Random());
+        if (cmdParam.repeatRun == true) {
+          // record the tests and the order in which they are run
+          BufferedWriter writer = new BufferedWriter(new FileWriter(new File(
```
cchang738: Same here, reuse "file".

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -187,7 +189,47 @@ public int runTests() throws Exception {
...
 
       LOG.info("> RUNNING TESTS (ITERATION " + i + ")..");
-      Collections.shuffle(tests, new Random());
+      String queryLogFilename = drillOutputDir + "/queryLog.log";
+      File file = new File(queryLogFilename);
+      if ((cmdParam.repeatRun == true) && (file.exists() && (file.length() > 0))) {
+        // run the tests in the requested order
+        try (BufferedReader reader = new BufferedReader(new FileReader(new File(
+                                     queryLogFilename)))) {
+          List<DrillTest> tests_copy = Lists.newArrayList(tests);
```
cchang738: Here can we do the switch at the end after you build the list? No copy needed just reassign?

rhou1:
Hi Chun,

I addressed your comments.  Please take a look and let me know if I missed something.
I also added some code to inform the user when a test is in the queryLog file but not in the source or group.  In which case, the test is skipped.  Only tests that are in the source (-s option) and group (-g option) can be run at this time.

rhou1:
I also fixed a minor issue with Random Failures, where the string "Random Failures" appeared in the Results section even if there are no Random Failures.

Agirish:
Hey @rhou1 , can you please detail how this is different from current behavior and how we can guarantee the same execution order? 

To clarify, if we already store collected tests in a specified order (say lexicographic), then shouldn't this be deterministic? 

Also, how do we handle cases with concurrency? A given query can run in 5 seconds or in 10 seconds, depending on load and other characteristics. So even if it reads a query list in your case to determine the order, can we be sure it's respected when multiple queries are run in parallel?

-1 until then.

rhou1:
The current behavior is to shuffle the tests and execute them.  So each execution has a different order.

The last line in the commit says this change mainly helps when there is no concurrency.  It is for debugging.

Agirish:
Agreed on the specific usecase. Thanks for clarifying. 

With regard to implementation, curious to understand if we could just not shuffle the tests if this parameter is specified - instead sort them. Then we will not require a file to be written into and read from.

rhou1:
If you don't sort/shuffle the tests, then I will guess you run the tests in the same order each time.

The intent of this change is to allow you to run the tests in the order you want to run them in.  For example, when I was testing External Sort, I found a test passed by itself.  But sometimes it failed when I ran the whole set, depending on how the tests were shuffled.  So I wanted to capture the specific order of tests that caused the one test to fail.

Having the file allows you to specify only the tests you want to run.  Ideally, you can narrow it down to a handful of tests.

Agirish:
Cool, thanks again. +1

I think this serves a narrow use case, but if it's helpful then we should get it in.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -187,7 +189,55 @@ public int runTests() throws Exception {
...
+          // record the tests and the order in which they are run
+          BufferedWriter writer = new BufferedWriter(new FileWriter(queryFile));
+          if (writer != null) {
+            for (DrillTest test : tests) {
+              writer.write (test.getInputFile() + "\n");
+            }
+            writer.close();
+          }
+        }
+      }
```
cchang738: Can you see if you can fit your code into the following logic structure:

if (cmdParam.repeatRun) {
  if (queryFile.length() > 0) {
    ...
  } else {
    ...
  }
}
original code...

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -276,14 +326,25 @@ public int runTests() throws Exception {
...
       }
+      if(skippedTests.size() > 0) {
+        LOG.info("Skipped Tests: These tests cannot be run from " + queryLogFilename + " because they were not in source or group");
+        for (String line : skippedTests) {
+          LOG.info(line);
+        }
+      }
       LOG.info(LINE_BREAK);
       LOG.info("Summary");
+      if ((cmdParam.repeatRun == true) && (queryFile.exists() && (queryFile.length() > 0))) {
```
cchang738: Can queryFile.length change? For example, if you started without the file but created afterwards, will this evaluate to true during the same run?

rhou1:
Hi Chun,

I updated the code based on your comments.  Please take a look.

rhou1:
Added try logic for BufferedWriter.  Chun, please take a look.

rhou1:
Reorganized the code a bit.  Chun, please take a look.

cchang738:
Did you rerun pre-commit test after the final revision? If yes, +1.

rhou1:
I made another change.  I no longer remove the query file from the "test" list at line 210.  This allows a test to be run more than once.

pre-commit test is here:
http://10.10.104.91:8080/job/Framework-Regression/1153/console

cchang738:
+1.

PR #330, opened 2017-10-10
MD-2882: Modified hive change metadata scripts

prasadns14:
Fixed the hive change metadata scripts, and the baseline accordingly. All test cases pass with hive 2.1 now.

@Agirish please review


framework/resources/Datasources/hive_storage/change_metadata2.ddl
```diff
@@ -34,45 +34,45 @@ ALTER TABLE remove_columns2 REPLACE COLUMNS (
...
   int_col float,
-  bigint_col int,
+  bigint_col bigint,
   date_col varchar(15),
-  time_col timestamp,
-  timestamp_col date,
-  interval_col boolean,
+  time_col string,
+  timestamp_col timestamp,
+  interval_col string,
```
Agirish: Should interval be string, or long?

Agirish:
+1 for 1.11.0-mapr. Before me merge into master, let's have another look to see if we are changing intended test behavior.

framework/resources/Datasources/hive_storage/change_metadata2.ddl
prasadns14: When the table was created the data type assigned to interval was string, retained the same. 
Table is created in the change_metadata1.ddl

Agirish: cool, thanks! @rchallapalli will be the best person to share more details on the intended behavior.

PR #331, opened 2017-10-10
MD-2893: Modified queries involving limit & their baseline

prasadns14:
Query drill3788.q - Table had multiple rows with same partkey, so also ordered by discount
Query 5level_sanity2.q - Table had 8 rows for each partkey, so reduced limit to 8

@Agirish please review


framework/resources/Functional/metadata_caching/partition_pruning/data/5level_sanity2.q
```diff
@@ -1 +1 @@
-select ps_supplycost from ps_5level where dir0=2000 and dir1='q1' and dir2='jan' and dir3=1 and dir4=1.0 or dir4=2.0 order by ps_partkey limit 10;
+select ps_supplycost from ps_5level where dir0=2000 and dir1='q1' and dir2='jan' and dir3=1 and dir4=1.0 or dir4=2.0 order by ps_partkey limit 8;
```
Agirish: I did not understand this change. Can you please explain why limit 10 was changed to limit 8?

Agirish:
+1, pending review comments.

framework/resources/Functional/metadata_caching/partition_pruning/data/5level_sanity2.q
prasadns14: The table has 8 rows per partkey. When sorted by partkey, the first 8 rows have partkey value 1 but the next two could be any of the 2 rows having partkey value 2. 
The ones being returned by GA branch are different from apache master.

Agirish: Can't we order by ps_supplycost as well?

prasadns14: Made the changes as suggested

Agirish: Thanks!

PR #334, opened 2017-10-11
MD-2916: Fixed baseline failures caused by column reordering

prasadns14:
Reordered columns to match the current behavior

@Agirish please review


Agirish:
+1. Thanks Prasad.

PR #335, opened 2017-10-11
MD-2919: Modified plan baseline

prasadns14:
Modified plan baseline to match the current behavior in GA branch.

Changes made - Modified baseline to have one file on each line & changed verification type to "regex-no-order"

@Agirish please review


Agirish:
@rhou1 can you also take a look?

Agirish:
These can be merged after @rhou1 reviews. +1 from my side.

rhou1:
+1 LGTM.  These tests are for single tables.

If there are multiple tables with multiple partitions, then I would have the "Scan" and "numFiles" parameter with each partition, so that partitions can be matched to each table.

PR #337, opened 2017-10-13
MD-2954: Fixed TPCDS100 hive datagen script

prasadns14:
Updated script to insert data into the tables.

We can't use external tables for all parquet files, as hive throws error for some parquet files (don't know the reason). So changed the script to create tables in those scenarios.

@Agirish please review


Agirish:
@prasadns14, can we figure out why those fail. It'll be good for consistency reasons. Also, we'd not want a huge data gen job to go through - instead of just pointing to a location via creating an external table

Agirish:
@prasadns14, is more work pending for this PR?

prasadns14:
@Agirish, all tests pass now

Agirish:
+1. Thanks Prasad. Will we need this on other branches?

PR #342, opened 2017-10-18
MD-3028: Create jdbc connection with/without plain authenticaion

prasadns14:
Create jdbc connection without username and password if password not set in drillTestConfig

@Agirish please review


Agirish:
You need a "then" after "if"

prasadns14:
@Agirish made the changes

Agirish:
+1. Thanks for making the change.

PR #352, opened 2017-10-24
DRILL-5795: Make test more robust for data addition in the plan

dprofeta:
DRILL-5795 updates the plan to add the number of rowgroup scanned
because the number of file scanned is less relevant. q28.e was
expecting the number of files and the useMetadata to be next to each
other. This is not the case anymore as the number of rowgroup is
in between. For robustness of the test, it is better not to assume they
are next to each other and in other tests, it is not expected.

This PR changes the expected string to a more robust one, so that it
works before and after the integration of DRILL-5795.

### DIFF
--- a/framework/resources/Functional/int96/q28.e
+++ b/framework/resources/Functional/int96/q28.e
@@ -1,1 +1,1 @@
-.*numFiles=2, usedMetadataFile=true.*
+.*numFiles=2.*numRowGroups=2.*usedMetadataFile=true.*


framework/resources/Functional/int96/q28.e
```diff
@@ -1 +1 @@
-.*numFiles=2, usedMetadataFile=true.*
+.*numFiles=2.*usedMetadataFile=true.*
```
cchang738: This is too relaxed. Need to verify numRowGroups as well. I already have a fix for this task. You can either change your PR or close it. I will post a PR.

cchang738: Or change yours to the following, thanks.

.*numFiles=2
.*numRowGroups=2
.*usedMetadataFile=true

dprofeta: I don't know what is your process for integration, but if I change to what your propose it means the PR in drill and in the test framework has to be merged at the same time. If one of them is merged and not the other, the test will fail.

cchang738: Yes, we merge in sync.

cchang738:
+1. Thanks.

PR #360, opened 2017-11-06
Updating data gen script for schema change tests on MapR-DB Binary

agirish:


Agirish:
@prasadns14, there was an issue with the script - on a test cluster, the loadTables script MR job failed with input path not found. I observed that the input path provided is a file, but is actually a directory. Updated accordingly. Can you please take a look?

prasadns14:
@Agirish, Since there were some issues with hbase.Import in 6.0, I changed the script to use hbase.ImportTsv instead. On the test cluster we should delete ${DRILL_TESTDATA}/schema_change_empty_batch/maprdb/binary/part, and also delete schema_change_empty_batch.tgz
This should fix the issue. I had made the changes in all the test clusters. Is this a new test cluster?

Agirish:
@prasadns14 , the script is using importTsv. But looks like it does not work as expected. There are multiple issues with the scripts - including copyFiles and loadTables. This should work on any test cluster, without requiring any manual steps. 

Looks like the changes i made are not complete. I'll update the PR and then we can take a futher look.

framework/resources/Datasources/schema_change_empty_batch/maprdb/binary/copyFiles.sh
```diff
@@ -5,10 +5,11 @@ set -x
 set -e
 
 if ! $(hadoop fs -test -d ${DRILL_TESTDATA}/schema_change_empty_batch/maprdb/binary/data)
-    then
+then
+        hadoop fs -rm -r -f ${DRILL_TESTDATA}/schema_change_empty_batch/maprdb/binary/data
```
prasadns14: Why delete when the folder is not there?

Agirish: Updated. I observed a case where the directory exists, but was empty. The script ran, but the subsequent script failed, since there were no input files. In my opinion, we should always remove and re-create the directory, to avoid such scenarios. Also cases were we update one or more files with the directory - they never take effect other wise.

Agirish:
Updated the PR to remove the changes to loadTables.sh (as they were resolved once the maven cache was cleaned up).

prasadns14:
LGTM +1

PR #363, opened 2017-11-06
MD-3142: Disabling test as it attempts to read column with >64K char"

agirish:


prasadns14:
Should we add this as a negative case rather?

Agirish:
Yes we could, once the Drill side changes are in.

PR #365, opened 2017-11-08
MD-3149: add drop table if exists

cchang738:
@prasadns14 Can you review this PR?

http://10.10.104.91:8080/job/Framework-Regression/1149/console


cchang738:
http://10.10.104.91:8080/job/Framework-Regression/1150/console

prasadns14:
Its a good improvement to increase the chances of cleanup. To actually know if the changes reduce random failures, we need to have a fair number of runs. Lets get this PR checked in to test it out.
LGTM, +1

rhou1:
Would we want to consider disabling the re-connection code if Dev wants to debug why the connection is dropped?

cchang738:
It should not affect any debug effort. Connection drop is still reported. Re-connection only happens for cleanup purpose.

PR #372, opened 2017-11-30
MD-2372: Removed overlap of hive scripts

prasadns14:
@Agirish, please review

Ran 5+ successful runs with pre commit launcher


Agirish:
Thanks a lot Prasad. This is going to be very useful, as the total runtime should significantly reduce. +1. LGTM

PR #383, opened 2017-12-19
Add new expected results files for Simba JDBC driver.

rhou1:
These are equivalent to the expected results files for the Apache JDBC driver.
The main differences are in formatting.
There are also a few minor plan changes.  Simba may add a ComplexToJson operator, which
means that subsequent operator numbers are increased by 1.


rhou1:
Krystal, please review these expected results files.

krystaln:
LGTM +1

PR #385, opened 2017-12-19
Add new expected results files for the Simba JDBC driver.

rhou1:
These are empty results files for tests that currently fail with the Simba JDBC driver.
They end in ".fail" so they are ignored by the test framework.  When these tests are fixed,
the ".fail" can be removed.


rhou1:
Krystal, please review these expected results files.

krystaln:
LGTM +1

rhou1:
Krystal, thank you for the review.

PR #386, opened 2017-12-22
Test Framework changes to incorporate Simba ODBC Driver into test automation framework.

rhou1:
Add new python program to use ODBC driver to execute SQL statements.  This python program creates an output file, as well as two metadata files to store the types of the columns (*.type file) and the names of the columns (*.label).
Adjusted DrillTestOdbc.java to execute a query file with multiple SQL statements.
Updated getTestCaseModeler to change submitType to "odbc" so that the tests are executed using the ODBC driver.  Also store the  python program in the script field.
Added progress counter that was introduced for the JDBC driver.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/DrillTestOdbc.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/DrillTestOdbc.java
@@ -22,10 +22,13 @@ import com.google.common.collect.Lists;
 import org.apache.drill.test.framework.TestCaseModeler.TestMatrix;
 import org.apache.drill.test.framework.TestVerifier.TestStatus;
 import org.apache.drill.test.framework.TestVerifier.VerificationException;
+import org.apache.drill.test.framework.TestVerifier.PlanVerificationException;
 import org.apache.log4j.Logger;
 
+import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.sql.Connection;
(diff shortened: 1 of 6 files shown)


rhou1:
Chun,  please review this when you can.  Krystal, please review the python program, which is similar to your perl script.

krystaln:
LGTM +1 on the python script.

framework/src/main/java/org/apache/drill/test/framework/DrillTestOdbc.java
```diff
@@ -71,20 +85,34 @@ public void run() {
...
-	LOG.info("Running test " + command);
+    LOG.info("Running test " + command);
     try {
 
   	  cmdConsOut = Utils.execCmd(command);
-  	  LOG.info(cmdConsOut.consoleOut);
+          LOG.info(cmdConsOut.consoleOut);
 
+      queries = Utils.getSqlStatements(modeler.queryFilename);
+      mainQueryIndex = queries.length / 2; // Currently, the main query must be in the middle of the list of queries
```
cchang738: Can we not have this restriction?

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -313,7 +313,16 @@ private static TestCaseModeler getTestCaseModeler(String testDefFile)
 	      throws IOException {
 	    byte[] jsonData = Files.readAllBytes(Paths.get(testDefFile));
 	    ObjectMapper objectMapper = new ObjectMapper();
-	    return objectMapper.readValue(new String(jsonData), TestCaseModeler.class);
+            // check if submitType should be over-ridden to odbc
+            TestCaseModeler result = objectMapper.readValue(new String(jsonData), TestCaseModeler.class);
```
cchang738: Can you make the variable "result" more descriptive?

framework/src/main/java/org/apache/drill/test/framework/DrillTestOdbc.java
Agirish: @cchang738 , you mean we could use the @test notation instead right?

rhou1:
Updated code to address review comments.

rhou1:
Updated the code to pull in another change from DrillTestJdbc to DrillTestOdbc.

cchang738:
LGTM. +1
Can you make a final run with the PR and post run summary here? Thanks.

rhou1:
Here is the run summary.

Passing tests: 8773
Execution Failures: 357
Data Verification Failures: 366
Plan Verification Failures: 137
Timeouts: 0
Canceled: 0
Random Failures: 0

PR #387, opened 2017-12-28
DRILL-5772: Check UTF-8 support set by default via saffron.properties…

omalashevsky:
… file


framework/resources/Functional/values/saffron_encoding_DRILL_5772.q
```diff
@@ -0,0 +1,5 @@
+drop table if exists encoding_test;
+create table encoding_test as select name, alias, description from (values('Василий Теркин', 'Вася', 'персонаж'), ('王維', '摩詰', '詩人'), ('Ибн Рушд', 'ابن رشد', 'философ'), ('James Cook', 'James', 'navigator')) as t(name, alias, description);
+select name from encoding_test where alias in ('Вася', '摩詰', 'ابن رشد');
+drop table if exists encoding_test;
+select 1 from (values(0));
```
prasadns14: @omalashevsky, why do we need the last select statement?

prasadns14: Did you run pre-commit tests?

cchang738: @omalashevsky You can mark the query that you want to verify result with "--@test". Doing it this way, you don't need the last query. I believe the README file explains it.  After correction, and rerun pre-commit, if it passes, you can merge the PR. +1. Thanks.

omalashevsky:
@Chun Chang
I have changed the test to use @test annotation. Thanks for pointing out. 
Regarding Pre-commit tests run, since CV QA team currently does not have test cluster, I have run only tests from values test suite. Is it possible to run Pre-commit tests on my branch in SJ before the commit?

cchang738:
@omalashevsky  I launched a pre-commit run for you. Will see how it goes.

PR #390, opened 2018-01-03
Add new expected results files for the Simba ODBC driver.

rhou1:
These are empty results files for tests that currently fail with the Simba ODBC driver.
They end in ".fail" so they are ignored by the test framework.  When these tests are fixed,
the ".fail" can be removed.


rhou1:
Chun,  please review these tests.

cchang738:
I don't feel it serves a good purpose checking in 800+ empty files just to skip tests, especially these tests are expected to be fixed in a month or two, and these files will be removed. If we only have a few failed tests, then this approach would be acceptable. I almost feel we might as well just leave these tests there and report them as failure if we want to check them in now. Or we can wait a bit until we get most tests passing before check them in. Another approach is to just have one failed_test_file listing all the currently failed tests to skip. This will need some test framework change but should be similar to code added to handle the failed expected files. I will let others to comment. It's not critical to me so I am fine either way.

rhou1:
I agree that 800+ empty tests is a lot.  But it is temporary.

I am not sure it is useful to implement a new way to handle failed tests if it is only needed for a few months.

PR #391, opened 2018-01-04
Enhance Test Framework to be able to run a single test.

rhou1:
The -s option can now support a mixture of test suites and individual query files,
separated by commas.  If the individual query file does not have a json definition
file in the same directory, the Test Framework will check the parent directories
until it finds at least one JSON file, at which time it assumes that one of those
JSON files can be used.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/Utils.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/Utils.java
@@ -195,6 +195,12 @@ public class Utils implements DrillDefaults {
 	  	LOG.error("Directory " + testDefSourceFile.getAbsolutePath() + " does not exist!");
     	  	System.exit(-1);
       }
+      if (testDefSourceFile.isFile()) {
+        // Single test execution
+        singleTestCase(testDefSourceFile, testGroups, drillTestCases);
+        continue;
+      }
+
       List<File> testDefFiles = searchFiles(testDefSourceFile, ".*.json");
       for (File testDefFile : testDefFiles) {
 //        try {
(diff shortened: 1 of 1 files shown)


rhou1:
Chun, please take a look.

Agirish:
Hey this is really cool! Thanks for adding this

Agirish:
Does this handle cases when there are more than one test def JSON files in the folder?

rhou1:
Yes, this supports having multiple JSON definition files in the same folder.  This is how I discovered the query file extension is not quite correct, because I was picking the wrong JSON file.

Agirish:
cool, thanks! I'd like to try this out sometime.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -747,4 +753,60 @@ public static String removeNewLines(String input) {
...
 
+  /* Get DrillTestCase for a single query file */
+  public static void singleTestCase(File singleTestFile,
+                                      String[] testGroups,
+                                      List<DrillTestCase> drillTestCases)
+      throws IOException {
+    List<File> jsonFiles = searchFiles(singleTestFile.getParentFile(), ".*.json");
+    if (jsonFiles.isEmpty()) {
+      String updir = "";
+      while (jsonFiles.isEmpty()) {
```
cchang738: What happens if you never encounter a test definition file in any of the upstream directories?

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -747,4 +753,60 @@ public static String removeNewLines(String input) {
...
+        }
+        if (skipSuite) {continue;}
+        String expectedFileName = getExpectedFile(singleTestFile.getAbsolutePath(),
+                  queryFileExtension, expectedFileExtension);
+        drillTestCases.add(new DrillTestCase(modeler, singleTestFile.getAbsolutePath(), expectedFileName));
+        break;
+      }
+    }
+  }
+
```
cchang738: Can you test the following scenarios:
1. -s .../dir contains test definition file and sudirs containing .json
2. -s .../test_definition_file.json
3. -s .../test_query.q
4. -s .../dir without any test definition files in any of the dir and parent dirs

Agirish: Do we need to specify the relative path to the test, including the file extension?

rhou1: What does the "..." do?

1. how is .json different from test definition file?
2. are there any files besides test_definition_file.json in the directory?

rhou1: @Agirish.  Yes.  Previously we would say
   -s Functional/p1tests
Now we can say
   -s Functional/p1tests/is_null.q

PR #393, opened 2018-01-04
DRILL-5166: Added tests to verify that NPE will not happen when query…

agozhiy:
…ing text files using the table function


framework/resources/Datasources/table_function/DRILL-5166_generate_data.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+source conf/drillTestConfig.properties
+
+test_dir=${DRILL_TEST_DATA_DIR}/Datasources/table_function
+dfs_location=/drill/testdata/table_function
```
cchang738: Should use ${DRILL_TESTDATA} here.

framework/resources/Datasources/table_function/DRILL-5166_generate_data.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
cchang738: Do we generate data every time we run automation? How big is the dataset and how long it takes to create the dataset? We want to reduce time spend in datagen phase. Please look into if we can generate data once and reuse in subsequent runs. Thanks.

framework/resources/Functional/table_function/positive/table_function.json
```diff
@@ -21,6 +21,11 @@
          "mode": "cp",
          "src": "Datasources/table_function",
          "dest": "table_function/"
+      },
+      {
+        "mode": "gen",
+        "src": "Datasources/table_function/DRILL-5166_generate_data.sh",
```
cchang738: Looks like your script copies generated file to dfs. But would the section "mode": "cp" does the same if your script put the generated file in the same "src" dir?

framework/resources/Datasources/table_function/DRILL-5166_generate_data.sh
Agirish: I agree. Please share how big the dataset is - depending on it, we could check it in to the repo or upload to S3.

framework/resources/Functional/table_function/positive/drill-5166_npe_with_table_function_group_select_plan_check.e
```diff
@@ -0,0 +1 @@
+"pop" : "union-exchange"
```
Agirish: For explain plan queries, we look for the text plan and not the JSON part. Please update accordingly.

Agirish: Also, please seperate out the data and plan verification tests - use a different extension (.plan) and use an additional JSON test def file. The verification type must be changed to regex.

framework/resources/Datasources/table_function/DRILL-5166_generate_data.sh
agozhiy: We can reuse it, good point.
Regarding to the size, I need the file larger than the maprfs chunk size, wich is 256MB by default. However I noticed that the Test Framework creates directories with chunk size 64MB and therefore I can reduce the file size to ~70MB.

framework/resources/Functional/table_function/positive/table_function.json
agozhiy: "mode": "cp" is executed prior to the "mode": "gen", so the generated files won't be copied by the "cp" section at the same test execution.

agozhiy:
Added changes according to your comments (commit e531831d9f56ed12c134d69b954ecd3343c0257e)
Please review.
Thanks!

framework/resources/Functional/table_function/positive/table_function.json
cchang738: True. If we pre-generate the data and reuse it, then we can rely on "cp" mode. The generated data needs to be put on s3. Let us know if you need help on that. This is our current practice how we handle large dataset. Thanks.

framework/resources/Functional/table_function/positive/plan/table_function_plan_check.json
```diff
@@ -0,0 +1,31 @@
+{
+  "testId": "table_function-functional",
```
Agirish: Please update testId to make sure it's unique (you have the same for both data and plan tests).

framework/resources/Functional/table_function/positive/plan/drill-5166_npe_with_table_function_group_select_plan_check.e
```diff
@@ -0,0 +1 @@
+UnionExchange
```
Agirish: Why are we only checking for UnionExchange in all tests?

framework/resources/Datasources/table_function/DRILL-5166_generate_data.sh
Agirish: Cool, then lets upload it to S3 (if you do not have an account, share the data over OneDrive and I'll upload it). Once it's available, i'll share the URL and you can add it to pom.xml, and update the tests to use the downloaded data instead. We'll no longer need this script then.

framework/resources/Functional/table_function/positive/plan/drill-5166_npe_with_table_function_group_select_plan_check.e
agozhiy: I need to check that more than one major fragments are being created. That was the case when the null pointer exception happened on previous releases. With only one fragment other tests will definitely pass but they will verify nothing.

Agirish: Got it, thanks.

agozhiy:
Changed testID to separate data and plan tests.
Also the test data is now compressed and the gen script untar that data if needed.
Please review.
Thanks!

Agirish:
Thanks for making the changes! I'll run this through regression and have this merged.

Agirish:
Run went through fine. Merging.

PR #404, opened 2018-01-12
DRILL-6028: UNION/-ALL queries on 1200 columns must be processed succ…

omalashevsky:
…essfully


Agirish:
Run went through fine. Merging.

PR #414, opened 2018-01-25
Updating TPC-DS Scripts

agirish:


Agirish:
@prasadns14 , can you please take a look?

prasadns14:
@Agirish do you think we can use su $user irrespective of whether the user is root or otherwise?

Agirish:
That's a good idea. Let me think and get back

Agirish:
That doesn't seem to work. We need to handle root user differently.

prasadns14:
+1

PR #423, opened 2018-02-12
MD-3538: Enabling test runs with security enabled

prasadns14:
@Agirish, @rhou1, @kkhatua please review


framework/src/main/java/org/apache/drill/test/framework/ConnectionPool.java
```diff
@@ -18,70 +18,57 @@
...
-    } else if (jdbcDriver.equals("com.mapr.drill.jdbc41.Driver")) {
-      TestDriver.driverType = TestDriver.DriverType.SIMBA_JDBC;
-    }
   }
 
   public Connection getOrCreateConnection(TestCaseModeler test) throws SQLException {
     final String username = test.matrices.get(0).username;
     final String password = test.matrices.get(0).password;
-    return getOrCreateConnection(username, password);
+    return getOrCreateConnection(username+password);
```
kkhatua: You might want to do something more than just concat. I've seen mapr/mapr login and also mapr/maprmapr , so this could be risky. I take it that we usually keep the user and password the same for test environments.. like kuser1/kuser1 , right?

prasadns14: Changed key to be username+"|"+password

conf/drillTestConfig.properties.example
```diff
@@ -86,6 +86,37 @@
...
+
+########################################
+# Truststore
+########################################
+#TRUSTSTORE_PATH=/opt/mapr/conf/ssl_truststore
+
+########################################
+# Truststore Password
+########################################
+TRUSTSTORE_PASSWORD=mapr123
```
Agirish: Some are commented and some aren't. Can you kindly double check and see if they can all be commented out, perhaps?

conf/drillTestConfig.properties.example
```diff
@@ -86,6 +86,37 @@
...
+
+########################################
+# Kerberos Principal
+########################################
+KERBEROS_PRINCIPAL=
```
Agirish: Do we expect a default value or an example here?

framework/pom.xml
```diff
@@ -32,7 +32,7 @@
       <drill.version>${env.DRILL_VERSION}</drill.version>
       <hadoop.version>2.7.1</hadoop.version>
       <runtimeDepenencies>${env.DRILL_HOME}/conf</runtimeDepenencies>
-      <apache-httpcomponents.version>4.1.1</apache-httpcomponents.version>
+      <apache-httpcomponents.version>4.5.5</apache-httpcomponents.version>
```
Agirish: This was for enabling REST over SSL right? It's good we are bumping it up.

framework/src/main/java/org/apache/drill/test/framework/DrillTestDefaults.java
```diff
@@ -0,0 +1,220 @@
...
+import com.google.common.collect.Maps;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.Map;
+import java.util.PropertyResourceBundle;
+import java.util.ResourceBundle;
+
+public class DrillTestDefaults {
```
Agirish: This is better!

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -622,23 +639,23 @@ private static String getHttpResponseAsString(HttpResponse response) throws IOEx
...
             + ".  Using /tmp for drill output");
-        TestDriver.drillOutputDir = "/tmp";
+        DrillTestDefaults.DRILL_OUTPUT_DIR = "/tmp";
       }
     }
     int index = inputFileName.lastIndexOf('/');
     String queryName = inputFileName.substring(index + 1);
     queryName = queryName.split("\\.")[0];
-    String outputFileName = TestDriver.drillOutputDir + "/" + testId + "_" + queryName;
+    String outputFileName = DrillTestDefaults.DRILL_OUTPUT_DIR + "/" + testId + "_" + queryName;
```
Agirish: Thanks for cleaning this up.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -533,72 +577,45 @@ public static String getSqlResult(ResultSet resultSet) throws SQLException {
...
-      builder.append("http://" + ipAddress + ":8047/storage/" + pluginType);
-      HttpPost post = new HttpPost(builder.toString() + ".json");
-      post.setHeader("Content-Type", "application/json");
-      post.setEntity(new StringEntity(content));
-      DefaultHttpClient client = new DefaultHttpClient();
-      HttpResponse response = client.execute(post);
-      return isResponseSuccessful(response);
+  public static boolean postDrillStoragePlugin(String content, String pluginType) throws IOException {
+    StringBuilder builder = new StringBuilder();
+    builder.append(protocol + DrillTestDefaults.DRILL_STORAGE_PLUGIN_SERVER + ":8047/storage/" + pluginType + ".json");
```
Agirish: Nice.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -99,24 +109,67 @@
...
+  }
+
+  /**
+   * Rest server authentication when plain authentication is enabled
+   */
+  public static void authenticateHTTPClient() {
+    HttpPost post = new HttpPost(protocol + DrillTestDefaults.DRILL_STORAGE_PLUGIN_SERVER + ":8047/j_security_check");
+    post.setHeader("Content-Type", "application/x-www-form-urlencoded");
+    List<NameValuePair> postParameters = new ArrayList<>();
+    postParameters.add(new BasicNameValuePair("j_username", "mapr"));
```
Agirish: Can we remove "mapr" here?

framework/pom.xml
prasadns14: yes

framework/src/main/java/org/apache/drill/test/framework/Utils.java
prasadns14: done

conf/drillTestConfig.properties.example
prasadns14: done

conf/drillTestConfig.properties.example
prasadns14: done

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -564,35 +579,33 @@ public int runTests() throws Exception {
...
-      LOG.info(">> File: " + filename);
-      String pluginType = filename.substring(0, filename.indexOf('-'));
-      Utils.updateDrillStoragePlugin(templateFile.getAbsolutePath(),
-    		drillStoragePluginServer, pluginType, fsMode);
-      Thread.sleep(200);
+    String templatePath = DrillTestDefaults.CWD + "/conf/plugin-templates/";
+    Utils.updateDrillStoragePlugins(templatePath);
+
+    if (DrillTestDefaults.IS_SECURE_CLUSTER) {
+      templatePath = DrillTestDefaults.CWD + "/conf/hive-plugin-templates-secure/";
```
Agirish: Can this be made a generic secure dir, rather than specific to hive?

Agirish: Also, just to understand better, we first upload all templates in the original dir. And then depending on it's secure or unsecure, we additionally upload the templates in one of the dirs? Perhaps we could make it more generic & structured? Say separating the contents in the original dir into "common", "secure" & "unsecure". Common should contain all plugins which don't have to be modified depending on security. Secure should contain all templates which are to be secure. And same with unsecure.

prasadns14: Yes, I am using the second approach. 

The only templates which are affected by the cluster being secure or not is hive. So, I now have two sets of hive templates. First the common ones are uploaded and then either secure hive or unsecure hive

Agirish: I was suggesting we create those sub folders with all templates in one among them. And make this generic, rather than specific to hive.

prasadns14: done

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -99,24 +109,67 @@
...
-    for (final String key : bundle.keySet()) {
-      properties.put(key.trim(), bundle.getString(key).trim());
+    return client;
+  }
+
+  /**
+   * Rest server authentication when plain authentication is enabled
+   */
+  public static void authenticateHTTPClient() {
+    HttpPost post = new HttpPost(protocol + DrillTestDefaults.DRILL_STORAGE_PLUGIN_SERVER + ":8047/j_security_check");
```
Agirish: Is there a way to make the port number configurable? Perhaps introduce a new constant in DrillTestDefaults? That can be overridden in drillTestConfig.properties

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -533,72 +577,68 @@ public static String getSqlResult(ResultSet resultSet) throws SQLException {
...
-   *          IP address of node to update storage plugin for
-   * @param pluginType
-   *          type of plugin; e.g.: "dfs", "cp"
-   * @return content of the specified plugin
-   * @throws Exception
+   * Update storage plugins in directory
+   * @param templatePath directory having drill storage plugin templates
+   * @throws InterruptedException
    */
-  public static String getExistingDrillStoragePlugin(String ipAddress,
```
Agirish: Have we removed this method?

prasadns14: yeah we don't use it anywhere. i believe we wouldn't need it for our testing

framework/src/main/java/org/apache/drill/test/framework/Utils.java
prasadns14: can be done.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
Agirish: No, I think we do need it. It backs up any existing storage plugins before the run and then restores after the run. Can you check and put it back?

Agirish: Lets add it back. Since even @cchang738 remembers this functionality. We can fix it later, in case its broken.

Agirish:
+1. LGTM. Thanks Prasad. 
May be run through regression again and merge it ?

PR #424, opened 2018-02-14
DRILL-6118: Fixed tests due to changes that replaced WILDCARD with DY…

agozhiy:
…NAMIC_STAR in some cases

Need the following PR to be merged into the Drill master before this:
https://github.com/apache/drill/pull/1104


Agirish:
@agozhiy, can you please update once the PR is merged. We can run through tests and merge it then.

Agirish:
DRILL-6118 has been committed to apache/drill master. Will have this merged soon (after the test run is done).

Agirish:
Runs are having some issue with env. Merging this in, as there are ~16 failures observed and this will fix the baseline.

PR #430, opened 2018-03-02
MD-3773: Hive orc transactional cases

prasadns14:
@Agirish please review


Agirish:
+1. Thanks a lot for this Prasad. It's nice to have these new tests added. This will be a good reference when we decide to expand our test coverage for Hive Transactions.

PR #437, opened 2018-03-09
MD-3770: Added SSL unit test

prasadns14:
@Agirish, please review

### DIFF
--- a/framework/pom.xml
+++ b/framework/pom.xml
@@ -38,6 +38,11 @@
       <version>${log4j.version}</version>
     </dependency>
     <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+    </dependency>
+    <dependency>
       <groupId>org.json</groupId>
       <artifactId>json</artifactId>
       <version>20090211</version>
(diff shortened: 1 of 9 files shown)


framework/pom.xml
```diff
@@ -37,6 +37,12 @@
       <artifactId>apache-log4j-extras</artifactId>
       <version>${log4j.version}</version>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/junit/junit -->
```
Agirish: May be lets take the comment off?

bin/sanity_tests
```diff
@@ -0,0 +1,7 @@
+source conf/drillTestConfig.properties
```
Agirish: May be call the file run_jdbc_sanity_tests?

framework/pom.xml
prasadns14: done

bin/sanity_tests
prasadns14: done

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -156,7 +156,7 @@ public void run() {
       }
       duration = stopwatch;
 
-      if (DrillTestDefaults.JDBC_DRIVER.equals(DrillTestDefaults.Driver.APACHE)) {
+      if (DrillTestDefaults.JDBC_DRIVER.equals("org.apache.drill.jdbc.Driver")) {
```
Agirish: How about we create a constant called APACHE_DRIVER in DrillTestDefaults.java and use that in all comparisons?

Agirish:
+1. Thanks a lot for working on this, Prasad. There is a lot we can do in the future to expand this further.

PR #439, opened 2018-03-21
DRILL-6118: Added tests and data for the verification that the item o…

agozhiy:
…perator with star is supported in cases of filter/project pushdown and partition pruning.


framework/resources/Functional/filter/pushdown/data/drill_6118_filter_push_down_with_view.sql
```diff
@@ -0,0 +1 @@
+select * from dfs.drillTestDir.`DRILL_6118_parquet_partitioned_by_folders_view` where c1>2;
```
cchang738: Without order by, will result set always the same? For small dataset, it should. But for large dataset, it might not be.

cchang738:
Did you run pre-commit automation test?

framework/resources/Functional/filter/pushdown/data/drill_6118_filter_push_down_with_view.sql
agozhiy: Order is not essential for these tests. As I know, Drill Test Framework ignores it unless the query contain "order by" clause. So the tests will fail only if the queries will return wrong data. I verified this case, intentionally changing result order, and the tests passed.

agozhiy:
@cchang738,  I didn't setup our test cluster for the pre-commit runs yet, but I run filter pushdown and partition pruning test suites on my VM, all tests passed.

cchang738:
+1. Thanks.

PR #443, opened 2018-03-27
Updating scripts invoking mapr shell

agirish:


Agirish:
While invoking `mapr dbshell`, we'd run into issues when executing the script as different users. Updating it by specifying the full path (`/usr/bin/mapr`) resolves the issue.

Agirish:
@kkhatua can you please take a look? For now, we'll merge this into 1.13.0-mapr branch and later cherry-pick into master.

Agirish:
This was the script which I mentioned was failing and blocking Advanced tests from running.

framework/resources/Datasources/tpcds/maprdb/json/importTables.sh
```diff
@@ -10,138 +10,138 @@ if [ -n $USERNAME ]
 then
     user=$USERNAME
 else
-    user=`ps -aef | grep Drillbit | grep org.apache.drill | cut -d' ' -f1 | head -1`
+    user=ps -aef | grep Drillbit | grep org.apache.drill | cut -d' ' -f1 | head -1
```
kkhatua: Don't you need the escape tics? Without that, the shell interpreter will assume the words after `ps` is a command.

framework/resources/Datasources/tpcds/maprdb/json/importTables.sh
```diff
@@ -10,138 +10,138 @@ if [ -n $USERNAME ]
...
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/web_site
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/store_returns
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/web_page
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/catalog_page
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/inventory
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/catalog_returns
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/web_returns
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/web_sales
-mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/catalog_sales
+/usr/bin/mapr dbshell drop /drill/testdata/tpcds_sf$sf/maprdb/json/customer
```
kkhatua: If you're needing to specify the `/usr/bin` explicitly I'm guessing that the environment is broken, because that directory should be part of `$PATH`

Agirish: The issue is not with script not being found - it was and the script would error out. The MapR script has an issue in my opinion - it does not work well when invoked in some situations.

framework/resources/Datasources/tpcds/maprdb/json/importTables.sh
Agirish: You are right, will fix it.

Agirish:
@kkhatua, i've addressed your comments. Please review.

kkhatua:
If there is a bug in the `mapr dbshell` script, as a consumer, we should log a bug. 
LGTM.
If the changes test out fine in a dry run, +1

Agirish:
I'll confirm if it's a bug (likely) and file one. Yes, the runs were clean.

PR #449, opened 2018-04-04
MD-3686: add nested loop test cases

cchang738:
Add test coverage for nested loop join with multiple batches from both sides of the join. Use multiple files to create multiple batches. Our pre-commit cluster is down so I ran locally and passed.

@krystaln please review. Thanks.


krystaln:
LGTM +1 as long as the pre-commit test is clean.

PR #451, opened 2018-04-12
Fix output for decimal values

vvysotskyi:
- Output before the changes:
These rows are not expected (first 10):
I��

- Output after the changes:
These rows are not expected (first 10):
-1E-29

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/Utils.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/Utils.java
@@ -22,6 +22,7 @@ import org.apache.commons.io.FilenameUtils;
 import java.io.*;
 import java.lang.reflect.Field;
 import java.net.*;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


vvysotskyi:
@Agirish, @kkhatua Could you please take a look?

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -538,7 +520,39 @@ public static String getSqlResult(ResultSet resultSet) throws SQLException {
...
+    int columnCount = resultSet.getMetaData().getColumnCount();
+    List<Object> values = new ArrayList<>();
+    for (int i = 1; i <= columnCount; i++) {
+      try {
+        switch (resultSet.getMetaData().getColumnType(i)) {
+          case Types.BINARY:
+          case Types.VARBINARY:
+            byte[] bytes = resultSet.getBytes(i);
+            if (bytes != null) {
+              values.add(new String(bytes, StandardCharsets.UTF_8.name()));
```
kkhatua: Don't we need to handle UTF-16 character set and Date objects?

Agirish:
@cchang738, can you also kindly take a look?

framework/src/main/java/org/apache/drill/test/framework/Utils.java
cchang738: This looks good to me and good improvement. Make sure you run all tests locally to confirm.
+1

vvysotskyi: @cchang738, it was checked with pre-commit tests.

Agirish:
+1. Looks good; thanks for making these changes, @vvysotskyi 
I'll see if I can run this through tests and will have this merged.

PR #453, opened 2018-04-18
Reduced logging and other improvements to console output

kedarbcs16:


Agirish:
+1. Please run through pre-commit tests and update if no issues.

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -39,6 +39,8 @@
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+
+
```
Agirish: Why do we have 2 new lines inserted here?

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
```diff
@@ -162,10 +165,8 @@ public void run() {
         LOG.info("[" + testStatus + "] (" + stopwatch + ") " + modeler.queryFilename + " (ConnectionID: " + connection.hashCode() + ")");
       }
 
-      if((++countTestsCompleted %100==0 && countTestsCompleted <= totalCases) || (countTestsCompleted == totalCases)){
-        LOG.info(LINE_BREAK);
-        LOG.info("Execution completed for " + countTestsCompleted + " (out of " + totalCases + ") tests");
-        LOG.info(LINE_BREAK);
+      if((++countTestsCompleted %100==0 && countTestsCompleted <= (totalCases*TestDriver.cmdParam.iterations*TestDriver.cmdParam.clones))){
```
Agirish: Isn't countTestsCompleted per iteration? If not it should be. That way we can keep the check as is.

framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
```diff
@@ -697,6 +698,8 @@ public TestStatus verifyTextPlan(String expectedOutput,
     if (verified) {
       return TestStatus.PASS;
     }
+    actual = actual.substring(0,ALLOWED_ACTUAL_PLAN_VERIFICATION_LENGTH);
```
Agirish: I think we should revert this. We need an intelligent way to crop the text.

conf/log4j.properties
```diff
@@ -1,6 +1,6 @@
 # Root logger option
-log4j.rootLogger=INFO, file, stdout
- 
+log4j.rootLogger=ERROR, file, stdout
+log4j.logger.DrillTestLogger=INFO, Drill
```
Agirish: Curious - why do we have "Drill" here?

framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
kedarbcs16: ok

framework/src/main/java/org/apache/drill/test/framework/DrillTestJdbc.java
kedarbcs16: it is not per iteration ,we have to change the logic for it ,this change is due to the fact that it is not an iteration specific thing thus ,the "Execution completed for  ..." is not printed for subsequent iterations

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -603,14 +603,16 @@ public static String getExistingDrillStoragePlugin(String ipAddress,
...
+    LOG.info("\n>> Path: " + templatePath + "\n");
     File[] templateFiles = new File(templatePath).listFiles();
     for (File templateFile : templateFiles) {
       String filename = templateFile.getName();
       LOG.info(">> Updating File: " + filename);
       String pluginType = filename.substring(0, filename.indexOf('-'));
       boolean isSuccess = Utils.updateDrillStoragePlugin(templateFile.getAbsolutePath(), pluginType, DrillTestDefaults.FS_MODE);
-      LOG.info(">> Update file " + filename + (isSuccess ? " succeeded" : " failed"));
+      if(!isSuccess){
+        LOG.info(">> Update file " + filename + (isSuccess ? " succeeded" : " failed"));
```
Agirish: if !isSuccess is true (upload fails), then there is no need for another check while printing, right?

Agirish:
+1. Thanks for making the changes.

PR #455, opened 2018-04-18
Error in the unexpected Count in data verification failure

kedarbcs16:


Agirish:
@kedarbcs16 have you tested to make sure it works? 
+1, Looks good to me.

rhou1:
Do we know why the count is not the same as the size of the list?

rhou1:
LGTM.  +1

cchang738:
+1.

PR #459, opened 2018-05-03
MD-4256 Baseline updates for decimal data type enhancements

denysord88:
Need to be merged after this Drill PR:
https://github.com/apache/drill/pull/1232


vvysotskyi:
This pull request may be merged since apache/drill#1232 was merged into Apache Drill master

Agirish:
@denysord88, @vvysotskyi, we generate baselines from Postgres/Greenplum. I am a bit hesitant updating the baselines - as we might want to be doubly sure before we change the expected values. 

(1) Changing of precision w/o actual loss of precision. Looks fine. 
0E-20 -> 0.000000

(2) Changing of precision w/ actual loss of precision. Needs clarification.
33.3333333333333333 -> 33.333333
15.3357280489492097 -> 15.33572804970625

1.0000000000000002E-54 -> 0E-38
-8.1000000729E-31 -> +0E-24

(3) Changing queries -> Needs clarification
DECIMAL(4,2) -> DECIMAL(6,2)

vvysotskyi:
Each math operation with decimals which has a decimal result should provide a way to calculate expecting scale and precision for the result considering only the scale and precision of its arguments - type inference. Therefore for some operations, like a division, we may receive a value with the smaller number of digits, since for the boundary cases result should have the same number of digits after the decimal point.

So for (2) case, such precision was predicted using type inference rules.
As for the case with 15.33572804970625, the expected result initially had a double precision, but with these changes, it was calculated more precisely. 
This one 1.0000000000000002E-54 was checked with hive, and hive returned 0 for it (it always trims zeros after the point for some reasons).

(3) A value, which was casted had precision 6, therefore query was modified. Also, it is possible to modify expected result instead of the query for this case: 1064.21 -> 1064.00, but we supposed that correcting the query is more suitable.

Agirish:
Thanks for clarifying @vvysotskyi. I have a few questions, but I'll have this merged first.

PR #464, opened 2018-06-07
DRILL-6331: Added tests to verify that optimizations like filter push…

agozhiy:
… down work with native Hive reader.


framework/resources/Functional/hive/hive_optimizations/data/DRILL_6331_count_into_scan_pushdown_managed_table.sql
```diff
@@ -0,0 +1,3 @@
+set `store.hive.optimize_scan_with_native_readers` = true;
+--@test
```
Agirish: Don't we need to reset the option after the test? 
Also, please remove the @test annotation - lets use that only when we have multiple SQL queries within a given test file and only one of them needs to be tested for correctness.

agozhiy: Done.

agozhiy:
@Agirish, addressed your comment, could you review?

Agirish:
Thanks @agozhiy. I'll have it run through regression tests and have it merged.

Agirish:
+1. LGTM. Tests passed.

PR #468, opened 2018-06-15
Update before-run.sql

agirish:


before-run.sql
```diff
@@ -1,9 +1,11 @@
...
 alter system reset all;
 alter system set `planner.enable_decimal_data_type` = true;
 alter system set `new_view_default_permissions` = '777';
 alter system set `planner.enable_limit0_optimization` = true;
 alter system set `exec.errors.verbose` = true;
 alter system set `planner.memory.max_query_memory_per_node` = 10737418240;
 alter system set `drill.exec.hashagg.fallback.enabled` = true;
+alter system set `drill.exec.hashjoin.fallback.enabled` = true;
+select * from sys.options where status = 'CHANGED';
+select * from sys.version;
```
kkhatua: Looks like this and the fallback for HashJoin was the only new entries enabled. 
Not like it matters really, but the version and drillbit state info should (ideally) be first. That way, any errors/failures due to alterations don't come as a surprise if not supported.

Agirish: I made that change intentionally. It was previously the first statements to run. When going through the console output in a run, I found it useful to keep the select statements down - for readability purposes. It's been that way in other branches which are being used - made master in sync now. I can show it in person sometime.

PR #472, opened 2018-06-27
Updating run_tests with increased Java thread stack size

agirish:
Tests sometimes fail with java.lang.StackOverflowError, during plan verification tests. It's likely a consequence of Catastrophic Backtracking (https://www.regular-expressions.info/catastrophic.html)

A workaround to this issue is to bump up the thread stack size for the java app as follows:

java -Xss40m ...


kfaraaz:
@Agirish change looks good, I am assuming we have a clean test run. LGTM.

Agirish:
Thanks! Yes, I've tested this change. Also, it's already in use in other branches - making this change for master branch now.

PR #473, opened 2018-06-27
Update Datasources/tpcds/maprdb/json/importTables.sh to enable dbshell

rhou1:
command to work.  For some reason, su needs a different form of the command when dbshell is used compared to importtable.

Abhishek, please review this change.


Agirish:
Hey Robert, thanks for fixing these. Can you please update it to use sudo -u instead? It will not need the additional quotes then. Also, that will be consistent with other updates we've made recently. We'll also need to fix a few other scripts which have similar syntax - we can do that in a later PR. 

Ex: 
sudo -u $user <COMMAND>

rhou1:
Abhishek, I updated the script.  Please take a look.

Agirish:
+1. LGTM. If you have a clean regression run, we can have this merged.

PR #476, opened 2018-06-27
Create CONTRIBUTING.md

agirish:


rhou1:
I personally prefer seeing things in one place, at least the important things.  When people have to click around, they tend to miss things.  So I personally would prefer to see the install information on the main page.

What happened to the order-by comments?

Agirish:
No it's too long to be in README. 

Regarding the Order-by comments, we need to move it to a different place - it's too verbose and needs to be outside of the main README. I'm trying to figure out the right place for it. I'm thinking of a docs directory and put it in that.

rhou1:
We should also document what parameters to change in Drill in order to run the test framework.  Things like Direct Memory.  We can't ask Abhishek to tune each person's setup.  Khurram and i both ran into problems with not allocating enough Direct Memory.

rhou1:
The order-by documentation is verbose because testing order-by is complicated.  If not done properly, people will think they are validating order-by tests when they are not.  Especially for new QA testers.  I'm not saying it should be in the README, but if not put in a visible place, it may be ignored.  Testing SQL is not easy, and I did not realize what I was missing with order-by testing until I dug into it.

Agirish:
I agree - I wrote the initial doc with this in mind. But I've come to realize that we can do better. I'm not sure where you got the message that i'm removing stuff to be unhelpful. 

I'm thinking of keeping the primary README as an overview page. I think i'll move back some of the stuff back here - such as the usage details. We need to update the parameters as well (it's partly outdated).

Now things like Direct memory configuration is outside the purview of this repository. It requires basic understanding of Drill administration - we could add a note about it in a requirements section and point to Apache Drill docs.

Agirish:
Regarding the ORDER-BY documentation, I think it's a good thing to have details. My issue is with where to keep the documentation and to rewrite parts of it to be more concise and with better formatting.

Agirish:
Take a look at https://github.com/agirish/drill-test-framework/tree/install (work-in-progress) to see how this will look like once merged.

Agirish:
Folks, please take a look to see if any of you have other concerns. If not we can get this in soon.

rhou1:
Looks good.

I noticed a couple of typos.

In the Execute Tests section, the explanation for the -n option should have "20" to match the example.

On the Contributing page, the title "Structure of test definiton files" needs an "i" in definition.

rhou1:
The order by section looks much better now with proper tables.  Thanks for doing this.

PR #484, opened 2018-07-19
DRILL-6471 tests for casting numbers and strings to decimal with wron…

denysord88:
…g scale and precision


denysord88:
@Agirish could you please review this negative tests?

PR #485, opened 2018-07-19
DRILL-6472 Add negative tests for cast decimal with zero precision

denysord88:


denysord88:
@Agirish could you please review this negative tests?

PR #488, opened 2018-07-20
Moved decimal_negative folder with tests outside of another functiona…

denysord88:
…l test suite


denysord88:
Negative test suite was in folder with functional test suite. Functional tests failed because they tried to run negative test suite and got an errors.
I moved folder with negative tests outside of functional test suite

@Agirish could you please review and merge this fix of negative test suite?

PR #491, opened 2018-07-23
DRILL-6603: Added tests to verify that filter push down is not happen…

agozhiy:
…ing if the column has no stats.


Agirish:
Running Tests.

PR #493, opened 2018-08-02
MD-3579: Compare double values with float precision

kedarbcs16:
Co-authored-by: kedarbcs16 <kedar.behera@gmail.com>
Co-authored-by: agirish <agirish@apache.org>

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
@@ -30,18 +30,19 @@ import java.util.List;
 public class ColumnList {
   private final List<Object> values;
   private final List<Integer> types;
-  private final boolean Simba;
+  private final boolean simba;
   public static final String SIMBA_JDBC = "sjdbc";
+  private static final double MAX_DIFF_VALUE = 1.0E-6;
 
   public ColumnList(List<Integer> types, List<Object> values) {
     this.values = values;
     this.types = types;
     if (TestDriver.cmdParam.driverExt != null &&
         TestDriver.cmdParam.driverExt.equals(ColumnList.SIMBA_JDBC)) {
-      this.Simba = true;
+      this.simba = true;
     }
     else {
-      this.Simba = false;
+      this.simba = false;
     }
   }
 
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -30,18 +30,19 @@
 public class ColumnList {
   private final List<Object> values;
   private final List<Integer> types;
-  private final boolean Simba;
+  private final boolean simba;
```
Agirish: Let's remove these changes from this PR - can create a new PR if needed for it

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -147,17 +154,23 @@ private boolean compare(ColumnList o1, ColumnList o2) {
...
         switch (type) {
         case Types.FLOAT:
         case Types.REAL:
           float f1 = (Float) list1.get(i);
           float f2 = (Float) list2.get(i);
-          if ((f1 + f2) / 2 != 0) {
-            if (!(Math.abs((f1 - f2) / ((f1 + f2) / 2)) < 1.0E-6)) return false;
-          } else if (f1 != 0) {
-            return false;
+          if (f1 != f2) {
```
rhou1: You removed the check for a zero denominator.

Does this work if f1 is zero and f2 is negative?

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -30,18 +30,19 @@
...
     this.values = values;
     this.types = types;
     if (TestDriver.cmdParam.driverExt != null &&
         TestDriver.cmdParam.driverExt.equals(ColumnList.SIMBA_JDBC)) {
-      this.Simba = true;
+      this.simba = true;
     }
     else {
-      this.Simba = false;
+      this.simba = false;
```
rhou1: What is the benefit of changing the capitalization?

framework/src/main/java/org/apache/drill/test/framework/TestVerifier.java
```diff
@@ -51,6 +49,7 @@
 public class TestVerifier {
   private static final Logger LOG = Logger.getLogger("DrillTestLogger"); 
   private static final int MAX_MISMATCH_SIZE = 10;
+  private static final double MAX_DIFF_VALUE = 1.0E-6;
```
rhou1: Is this used?

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
kedarbcs16: the prev 0 check was redundant ,but valid point id does need a zero denominator check

PR #497, opened 2018-08-14
Added tests for Drill fragments with decimal datatype in big files

denysord88:


denysord88:

framework/resources/Functional/drill_fragments/decimal/data/group_by.sql
```diff
@@ -0,0 +1,12 @@
+SET planner.width.max_per_node=100;
+SET planner.slice_target=1;
+SET planner.enable_multiphase_agg=false;
+--test
```
Agirish: The '--test' is just a comment. Did you intend to put in the --@test annotation? Just FYI, it's not required unless you have uneven number of statements in the file. Here, since you have the middle query the actual test, you are good. 

But if say for example you had 4 statements and wanted to test the 3rd statement, you need to add the --@test annotation.

denysord88: Done. I removed "--test" lines from all tests

PR #498, opened 2018-08-21
MD-4707: Added tests to check that Drill can use JsonTableGroupScan f…

agozhiy:
…or Hive-MapR DB json tables


Agirish:
@kkhatua , I did take a look and it looks good. I have a few comments - will add them shortly.

conf/plugin-templates/common/dfs-storage-plugin.template
```diff
@@ -572,6 +572,9 @@
       ],
       "extractHeader": true,
       "delimiter": ","
+      },
+      "maprdb": {
```
Agirish: We use a different plugin for maprdb. I guess it's mfs or something. Can you use that instead?

framework/resources/Functional/hive/hive_storage/hive_maprdb_json/plan/hive_maprdb_json_plan.json
```diff
@@ -0,0 +1,35 @@
+{
```
Agirish: We don't need 'data' and 'plan' folders, since the extensions are different anyway. 
For data, it will pick up .sql and .e_tsv files
For plan, it will pick up .sql.explain and .plan files
Can you please combine them into the same dir?

framework/resources/Datasources/json_storage/MD_4707/MD_4707_import_tables.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+source conf/drillTestConfig.properties
+
+hadoop_folder=${DRILL_TESTDATA}/MD_4707
+test_data=MD_4707_test_data.json
+maprdb_table=MD_4707_json_MapR_DB_table
```
Agirish: Instead of the /MD_4707/MD_4707_json_MapR_DB_table dir, can we use something else more clear? Say hive/maprdb/json/<table_name> directory?

Also i would suggest you use tpch or tpcds dataset, that way we do not introduce another new dataset into the framework. All we need is to create a Hive external table for an existing tpch/tpcds maprdb table. I think I already have few in.

agozhiy: Usually I prefer to use my own tables if they don't take a lot of space. And re-using of data may result in unexpected failures, there were such incidents.

framework/resources/Functional/hive/hive_storage/hive_maprdb_json/plan/hive_maprdb_json_plan.json
agozhiy: Are you sure? I think it's better to have a single test suite per folder. They have a clearer look that way and easier to maintain.

framework/resources/Datasources/json_storage/MD_4707/MD_4707_import_tables.sh
Agirish: External tables should be safe. I want to eventually get rid of these small random datasets and work on standardizing on data sets and tests. So please update the tests to use the tpch/tpcds maprdb tables which are created in a different suite.

framework/resources/Functional/hive/hive_storage/hive_maprdb_json/plan/hive_maprdb_json_plan.json
Agirish: Yes, sure. That was the whole point of suggesting unique extensions. We are adding un-necessary levels of nesting directories otherwise. Also, it's easy to map a data query with a plan query, as they would now share the same name (with different extension). So we can have sets of 4 files per test which require both data and plan verification. It's a huge task to rework this later, so lets get this done when we are adding them.

agozhiy: Done.

framework/resources/Datasources/json_storage/MD_4707/MD_4707_import_tables.sh
agozhiy: Renamed the folder to hive/maprdb/json/. 
As for the tpch/tpcds maprdb tables, they all have column names in upper case and cannot be used with Hive due to the bug: https://maprdrill.atlassian.net/browse/HIVE-298. @Agirish  Could you suggest any workaround?

agozhiy:
@Agirish, I addressed your comments, could you please review?

Agirish:
Discussed with @agozhiy. There are a few minor changes requested, after which we can get this in for now. Eventually lets work on consolidating tests and datasets. 

Right now we have several thousand of tests and many of them are redundant and can be removed. And there are multiple others which can be combined  with other tests to make them more comprehensive & organized (like reusing similar tests for different data sources). 

Another issue is bloat - we have too many datasets and scripts. It becomes harder to maintain. We should try and reuse datasets as much as possible, and only introduce new ones when really needed.

PR #500, opened 2018-08-24
Set `store.hive.optimize_scan_with_native_readers` to true to prevent

rhou1:
some TPCDS tests from failing when using hive reader until HIVE-19069 is fixed.

### DIFF
--- a/framework/resources/Advanced/tpcds/tpcds_sf100/hive/parquet/query15.sql
+++ b/framework/resources/Advanced/tpcds/tpcds_sf100/hive/parquet/query15.sql
@@ -1,3 +1,5 @@
+-- set hive option until HIVE-19069 is fixed
+set `store.hive.parquet.optimize_scan_with_native_reader` = true;
 -- start query 15 in stream 0 using template query15.tpl 
 SELECT ca_zip, 
                Sum(cs_sales_price) 
@@ -18,3 +20,4 @@ WHERE  cs_bill_customer_sk = c_customer_sk
 GROUP  BY ca_zip 
 ORDER  BY ca_zip
 LIMIT 100; 
+reset `store.hive.parquet.optimize_scan_with_native_reader`;
--- a/framework/resources/Advanced/tpcds/tpcds_sf100/hive/parquet/query19.sql
+++ b/framework/resources/Advanced/tpcds/tpcds_sf100/hive/parquet/query19.sql
@@ -1,3 +1,5 @@
+-- set hive option until HIVE-19069 is fixed
+set `store.hive.parquet.optimize_scan_with_native_reader` = true;
 -- start query 19 in stream 0 using template query19.tpl 
 SELECT i_brand_id              brand_id, 
                i_brand                 brand, 
@@ -29,3 +31,4 @@ ORDER  BY ext_price DESC,
           i_manufact_id, 
           i_manufact
 LIMIT 100; 
+reset `store.hive.parquet.optimize_scan_with_native_reader`;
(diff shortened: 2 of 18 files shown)


rhou1:
Krystal, can you review these changes?  These tests are failing in Jenkins Advanced Test Baseline.

framework/resources/Advanced/tpcds/tpcds_sf100/hive/parquet/query15.sql
```diff
@@ -1,3 +1,5 @@
+-- set hive option until HIVE-19069 is fixed
+alter session set `store.hive.optimize_scan_with_native_readers` = true;
```
Agirish: Can you change this to just "set" - "alter session" is not needed. 
For example, "set `store.hive.optimize_scan_with_native_readers` = true;"
Similarly, just "reset `store.hive.optimize_scan_with_native_readers`;"

rhou1: I personally like seeing "session" vs. "system" options.

Why is using "set" better than "alter session"?

Agirish: That's the decision Drill made few years back - "set" is a short form for "alter session set". All tests have been since using the updated syntax. So lets make sure not to introduce inconsistency. All options used in tests are for session, so need for the redundancy. All system options need to be set in the before-run/after-run.sql

rhou1: We don't have consistency in our tests.  We recently had a discussion about explain plans, and how some people prefer multiple lines while others prefer single lines.  And I think it is good to have some flexibility so that engineers can write tests in a way that works for them.

kfaraaz: Agree with @rhou1  that we should only set (session) specific options inside tests. Setting them system wide using alter system, can affect other tests in the suite. Always, set options at a session level for a specific test, not at the system level, unless all other tests need that option set at a system level.

Agirish: @rhou1 , I disagree - having consistency always has benefits. When someone new has to work on maintaining these, having multiple inconsistent ways to doing things can become challenging. It's true that we do not have the consistency we need w.r.t the explain plan tests - it's a pain to fix them because of the bad implementation we have for the regex verification. That doesn't justify creating more mess. 

And there is consistency w.r.t the 'alter session set' vs 'set' - i've been making sure of it.

Agirish: @kfaraaz , you mean you are agreeing with me - I made those statements regarding session vs system :)

rhou1: Well, we will have to disagree that consistency *ALWAYS* has benefits.  People see things in different ways.  I can see how someone views a multiple line plan as more logical while someone else views a single line plan as more logical.  If it is easier for the author to understand a multiple line plan and verify their tests are correct, I can respect that.

Having said that, some consistency can be valuable.  It would be really nice if we could have a group decision on these issues.  Otherwise, someone is flipping a coin and everyone else has to live with it.  And sometimes choosing one "consistent" way to do things can require making a lot of unnecessary changes to old tests.

I will make the change here, since I want to get Jenkins clean.  But I think we can do better in the future.

arina-ielchiieva:
`store.hive.optimize_scan_with_native_readers` is deprecated, consider using `store.hive.parquet.optimize_scan_with_native_reader` instead.

Agirish:
Hey @rhou1 , is there anything pending in this PR?

rhou1:
Jenkins cluster was not available for testing at the time this PR was ready, hence the delay.  I just ran the Advanced tests and they pass.

PR #505, opened 2018-09-18
DRILL-6624: the same schema name specified as single path and as a co…

denysord88:
…mplex path in the same query


denysord88:

PR #509, opened 2018-10-18
MD-5037

kedarbcs16:
the commented logs are still there ,tested on my cluster .


framework/resources/Functional/amplab/amplab.json
```diff
@@ -12,7 +12,8 @@
             "schema": "dfs.drillTestDirAmplab",
             "output-format": "tsv",
             "expected-file": ".*.e_tsv",
-            "verification-type": [
+            "create-data-in-folder": true,
```
Agirish:
```suggestion
            "capture-output": true,
```

framework/resources/Functional/amplab/amplab.json
```diff
@@ -12,7 +12,8 @@
             "schema": "dfs.drillTestDirAmplab",
```
Agirish: I know this is an example, but I don't think we want to check this in - we aren't aiming to generate output files for Amplab tests.

framework/src/main/java/org/apache/drill/test/framework/CmdParam.java
```diff
@@ -16,7 +16,10 @@
     
     @Parameter(names = {"-d"}, description = "generate data", required=false)
     public boolean generate = false;
-    
+   
+    @Parameter(names = {"-o"}, description = "create baseline in same folder", required=false)
```
Agirish:
```suggestion
    @Parameter(names = {"-co"}, description = "capture output files as baselines", required=false)
```

framework/src/main/java/org/apache/drill/test/framework/CmdParam.java
```diff
@@ -16,7 +16,10 @@
...
+    public boolean createDataInFolder = false;
```
Agirish:
```suggestion
    public boolean captureOutputFiles = false;
```

Agirish:
@kedarbcs16 I'll close this PR. There is more work to be done. Let's discuss offline on the best way forward. We could then reopen this or create a new PR.

PR #511, opened 2018-11-01
MD-5018: Copy csv to dfs for every run

abhidotravi:
- The existing script does not copy `DRILL_5941_lines_count_skip_header_footer_big_data_set.csv` to dfs location if it already exists. It is possible that previous copy did not go through completely and we may have a corrupt data. This could result in tests failing.
- The fix is to copy csv every time.

### DIFF
--- a/framework/resources/Datasources/hive_storage/skipHeaderFooter/DRILL_5941_prepare_data.sh
+++ b/framework/resources/Datasources/hive_storage/skipHeaderFooter/DRILL_5941_prepare_data.sh
@@ -37,11 +37,7 @@ prepare_dataset () {
     local tar_name=$3
     local destination=$4
 
-    # Reusing of existing file if exists
-    if [ ! -f $file ]
-    then
-        untar_data $location $tar_name $file_name
-    fi
+    untar_data $location $tar_name $file_name
 
     copy_to_dfs $location $file_name $destination
 }


framework/resources/Datasources/hive_storage/skipHeaderFooter/DRILL_5941_prepare_data.sh
```diff
@@ -25,12 +25,9 @@ copy_to_dfs () {
...
     fi
-
-    hadoop fs -test -f $destination/$file_name
-    if [ ! $? -eq 0 ]
-    then
-	    hadoop fs -put $file $destination/
-    fi
+    
+    # Replace the csv every time
+    hadoop fs -put -f $file $destination/
```
kkhatua: Doesn't this increase the run time always? If it's a small increase, it is probably ok.  I'm wondering if there are better ways to check for potential corruption, like an MD5 checksum on a test dataset, though I'm not sure there is a hadoop command to do the same.

Agirish: Yeah, that's true. Here since it's a data gen scripts which is doing the copy, we anyway delete and re-create data - this is across scripts. This was wasn't doing that and hence the fix.

As part of test framework data copy functionality we could think of the checksum approach. 

Although I think overwriting is fine as these are done only on a nightly basis - so increase in total runtime won't impact most runs.

agozhiy:
For the consistent approach I'd recommend to remove this check as well:

```bash
    if [ ! -f $file ]
    then
        untar_data $location $tar_name $file_name
    fi
```

aravi5:
@agozhiy  - I have removed the check for untar and tested the changes locally.

agozhiy:
+1, Thank you for making the changes.

PR #517, opened 2018-11-27
Fixed the gen script for the the Hive MapR DB tables.

agozhiy:
Cannot create MapR DB tables as root, so now they are created with mapr user explicitly.

### DIFF
--- a/framework/resources/Datasources/hive_storage/hive_maprdb_json/import_tables.sh
+++ b/framework/resources/Datasources/hive_storage/hive_maprdb_json/import_tables.sh
@@ -1,30 +1,18 @@
 #!/usr/bin/env bash
 source conf/drillTestConfig.properties
 
-if [ -n $USERNAME ]
-then
-    user=$USERNAME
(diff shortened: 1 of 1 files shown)


agozhiy:
@Agirish, could you please review and check if it helps?

agozhiy:
The thing is, you cannot create a MapR DB table as a root:
```
Diagnostics: Application application_1543316577607_0035 initialization failed (exitCode=255) with output: Requested user root is not whitelisted and has id 0,which is below the minimum allowed 500

Failing this attempt. Failing the application.
```
So you need some other user to execute the gen script.
Possible solutions:
* Edit file /opt/mapr/hadoop/hadoop-2.7.0/etc/hadoop/container-executor.cfg and set min.user.id=0 (I think this is a bad idea)
* Don't run tests as root 
* Use a non-root user but parametrize it

But first I need to find out if my guess is right. @Agirish, could you please run a job with my branch to check if it fixes the tests?

PR #518, opened 2018-11-30
Removed newlines in gen scripts as Sqlline 1.6.0 handles them incorre…

agozhiy:
…ctly.


agozhiy:
@kkhatua, could someone please merge this? It fixes the crash during the data generating.

PR #521, opened 2018-12-10
DRILL-6834: Added possibility to run queries that don't return result…

agozhiy:
… set.

And option "exec.return_result_set_for_ddl" was introduced in Drill, when it is false, some queries don't return result set, that cause an SQL exception in Drill Test Framework.
To fix that, Statement execute() method now used instead of executeQuery(). Also added required null checks.


agozhiy:
Actually, changing this option to false will cause the issue I mentioned. So, for example, executing set/reset option queries will result in SQL exception and the tests will fail with execution failure.

Agirish:
I was asking about setting this to true and avoiding queries which return no result set. Is the goal to explicitly have tests which return no result set and so update framework to add support for it?

agozhiy:
Yes, I need to test that feature, and my tests are failing.

PR #523, opened 2018-12-13
DRILL-6834: Added tests to verify Drill option exec.return_result_set…

agozhiy:
…_for_ddl

### DIFF
--- a/framework/resources/Functional/no_result_set/DRILL_6834_select_query.sql
+++ b/framework/resources/Functional/no_result_set/DRILL_6834_select_query.sql
@@ -1,0 +1,3 @@
+set `exec.return_result_set_for_ddl` = false;
+select first_name, last_name from cp.`employee.json` order by employee_id limit 5;
+reset `exec.return_result_set_for_ddl`;
--- a/framework/resources/Functional/no_result_set/DRILL_6834_set_option_no_result_set.sql
+++ b/framework/resources/Functional/no_result_set/DRILL_6834_set_option_no_result_set.sql
@@ -1,0 +1,5 @@
+set `exec.return_result_set_for_ddl` = false;
+--@test
+set `store.format` = 'json';
+reset `store.format`;
+reset `exec.return_result_set_for_ddl`;
--- a/framework/resources/Functional/no_result_set/DRILL_6834_use_no_result_set.sql
+++ b/framework/resources/Functional/no_result_set/DRILL_6834_use_no_result_set.sql
@@ -1,0 +1,3 @@
+set `exec.return_result_set_for_ddl` = false;
+use dfs.`drillTestDir`;
+reset `exec.return_result_set_for_ddl`;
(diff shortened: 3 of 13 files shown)


framework/resources/Functional/no_result_set/DRILL_6834_set_option_no_result_set.sql
```diff
@@ -0,0 +1,5 @@
+set `exec.return_result_set_for_ddl` = false;
+--@test
+set `store.format` = 'json';
```
Agirish: Not sure about what this is testing. May be the CTAS is missing?

framework/resources/Functional/no_result_set/DRILL_6834_use_no_result_set.sql
```diff
@@ -0,0 +1,3 @@
+set `exec.return_result_set_for_ddl` = false;
+use dfs.`drillTestDir`;
```
Agirish: Same here. Not sure about what this is testing. May be the CTAS is missing?

framework/resources/Functional/no_result_set/DRILL_6834_select_query.sql
```diff
@@ -0,0 +1,3 @@
+set `exec.return_result_set_for_ddl` = false;
+select first_name, last_name from cp.`employee.json` order by employee_id limit 5;
```
Agirish: This is just a valid usecase right? The option is meant for DDLs. But sure, good to test if this works for all SQL statements.

Agirish:
Hey @agozhiy , some of the tests look incomplete - can you please take a look and comment?

framework/resources/Functional/no_result_set/DRILL_6834_set_option_no_result_set.sql
agozhiy: This is not a DDL statement, but it is also affected by the option tested.

framework/resources/Functional/no_result_set/DRILL_6834_use_no_result_set.sql
agozhiy: Please see my previous comment.

framework/resources/Functional/no_result_set/DRILL_6834_select_query.sql
agozhiy: Yes, you're right. We need to be sure that if the option is false, regular SQL queries will return result.

agozhiy:
@Agirish, I addressed your comments, please review.

framework/resources/Functional/no_result_set/DRILL_6834_set_option_no_result_set.sql
Agirish: So the SQL statement being tested here is the `set option=value`. Got it thanks.

PR #524, opened 2018-12-17
DRILL-6865: Added tests to verify the bug fix.

agozhiy:
Cases covered:
1. Filter pruning shouldn't work if at least one filter cannot be pruned ('like' operator or expression on a column).
2. Filter pruning should work in case of single parquet file with single row group.


framework/resources/Functional/filter/pushdown/filter_pruning/DRILL_6865_no_filter_pruning_expression_in_filter.plan
```diff
@@ -0,0 +1,2 @@
+Filter
```
Agirish: Are we sure to keep this as just "Filter" without regex? Like '.*'

Also here and in other tests, we should test the filter condition as well - like '< 5', plus columns being pushed down etc,

framework/resources/Functional/filter/pushdown/filter_pruning/plan.json
```diff
@@ -0,0 +1,26 @@
+{
+  "testId": "filter_pushdown_filter_pruning",
```
Agirish: I think we should make the test ids unique. I know the extensions help here, but if we want to test for duplicates later, these may show up as false negatives.

framework/resources/Functional/filter/pushdown/filter_pruning/data.json
```diff
@@ -0,0 +1,27 @@
...
+      "verification-type": [
+        "in-memory"
+      ]
+    }
+  ],
+  "datasources": [
+    {
+      "mode": "cp",
+      "src": "Datasources/parquet_storage/DRILL_3855/DRILL_3855_test_data",
+      "dest": "parquet_storage/DRILL_3855/DRILL_3855_test_data"
```
Agirish: do you think this can be simpifiled - like remove this sub-dir or rename it as just data, as we know it's under DRILL-3855?

framework/resources/Functional/filter/pushdown/filter_pruning/DRILL_6865_no_filter_pruning_expression_in_filter.plan
agozhiy: We don't need to match a whole line with an operator, so wildcard is used only inside the matching strings.
Added conditions and columns except one case, on which a new JIRA was submitted: DRILL-6910.

framework/resources/Functional/filter/pushdown/filter_pruning/plan.json
agozhiy: That was unintentionally. Fixed now.

framework/resources/Functional/filter/pushdown/filter_pruning/data.json
agozhiy: Your'e right. Removed the redundant sub-dir here and in other filter pushdown tests as well. I have a concern though: as the data is already on clusters there would be duplicates. Don't sure about the best way to solve this except to clean it manually.

framework/resources/Functional/filter/pushdown/filter_pruning/DRILL_6865_no_filter_pruning_expression_in_filter.plan
Agirish: Thanks for sharing!

framework/resources/Functional/filter/pushdown/filter_pruning/data.json
Agirish: Actually, I was talking about the sub dir. 

It’ll be good to have </drill/testdata>/<feature>/<JIRA>/<data>

When you do a hadoop ls, it’ll show consistent directories for JIRAs. Some may have one file and some may have many files.

Agirish:
@agozhiy , once you make the requested changes, we can have this merged.

framework/resources/Functional/filter/pushdown/filter_pruning/data.json
agozhiy: Moved the data to the </drill/testdata>/filter/pushdown folder. Most of it are Drill tables, which themselves represented as folders, so we don't need any intermediate directory. And so I left JIRA name as  part of the table names.

agozhiy:
@Agirish, I made the changes. Also, I squashed the commits into one and rebased to master.

PR #525, opened 2018-12-24
Fixed baselines due to the renamed option

agozhiy:
exec.return_result_set_for_ddl -> exec.query.return_result_set_for_ddl


arina-ielchiieva:
@Agirish please review. It's safe to rename option since it was introduced in 1.15 which it's not released yet.

PR #527, opened 2019-01-08
Fixed two advanced tests by casting doubles to decimals

agozhiy:


Agirish:
Looks good. But I'm concerned about changing the TPC-H original query to get this working. Any idea why this is specific to MapR-DB data source and not seen with Parquet or elsewhere?

agozhiy:
@Agirish, this is specific to the way how Java handles double values. Double is not precise as decimal, so even the order of calculations can affect the result. This shouldn't be MapR-DB specific so I think it is accidental. 
Casting to decimal is not the only possible solution. For example:
- You can modify the data to contain decimal values instead of double (I'm not sure if MapR-DB supports it though)
- Add order by to the subqueries, like `select sum(L_DISCOUNT) from (select L_DISCOUNT from mfs.tpch_sf1_maprdb_json.lineitem order by L_DISCOUNT)` to eliminate the influence of data processing order.
- Apply round to the result

Agirish:
Hey Anton, thanks for the explanation. Kedar and I are trying to see if we can enhance the test framework to handle some of these type of precision issues. Let's hold off merging this PR until that's done. Tentatively by end of this week.

agozhiy:
Closing this PR as the solution is unacceptable.

PR #528, opened 2019-01-08
Detect DRILL_VERSION from git.properties

kkhatua:
This would address issues arising out of 4 digit versioning of Drill or any other similar changes (e.g. version...-SNAPSHOT)

### DIFF
--- a/conf/drillTestConfig.properties.example
+++ b/conf/drillTestConfig.properties.example
@@ -146,7 +146,7 @@ export M2_HOME=/usr/local/apache-maven/apache-maven-3.1.1
 export M2=$M2_HOME/bin
 export PATH=.:$M2:$PATH
 
-export DRILL_VERSION=1.6.0-SNAPSHOT
+export DRILL_VERSION=$(grep 'git.build.version' ${DRILL_HOME}/git.properties | tr '=' '\n' | tail -1)
 export HADOOP_VERSION=1.0.3-mapr-3.0.2
 export USERNAME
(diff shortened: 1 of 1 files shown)


Agirish:
Hey Kunal, 

This is just a example file and is not actually used. In our test infrastructure, we have Jenkins jobs which already take care of this, so I don't think this change is necessary or consumed. 

Also, the process is a bit complicated, as we do not setup the test framework for each run - it's pre-setup for a specific version of Drill. I'll explain more in person.

kkhatua:
Yes, I noticed this is an example file.
The reason I put it in is because starting with MEP 6.1, the versions will change with almost every commit (atleast on the EBF branches), with the 4th digit getting a refresh/increment. We need to be able to detect that or atleast run without disruption.
This example file will tell anyone doing a new setup on how to manage this constantly changing property. Already made a change on the regression cluster (`10.10.104.61`) in the 1.15.0 config. [`/root/drillAutomation/1.15.0/conf/drillTestConfig.properties`]

Agirish:
Sure, we could get it in as a reference. But it will not work with RPM / DEB installations of Drill. Should work if Drill is built from source and deployed (git.properties isn't guaranteed to be present in former - @aravi5 helped point that out).

For internal, the config you made isn't persisted for long, it is overwritten on a nightly basis. I'll explain  more in person.

Agirish:
Merging this in, with caveat that it's just for reference, and works in most cases.

PR #529, opened 2019-01-14
DRILL-6783: CAST string literal as INTERVAL MONTH/YEAR works inconsis…

denysord88:
…tently when selecting from a table with multiple rows


denysord88:

framework/resources/Functional/explicit_cast/DRILL_6783_single_string_pmonth_as_interval_month.q
```diff
@@ -0,0 +1 @@
+select cast('P314M' as interval month) from cp.`employee.json` limit 10;
```
Agirish: I understand you are trying to cast a literal and are not using the employee.json file. Can you instead use the values operator? Just so that it's more clear to readers.

denysord88: Issue is reproducing on the result set with more then 4 lines. I can't use value operator here.

Agirish: Thanks for confirming. Yes, I tried with values operator, and getting more rows will be a challenge. We could use sys.options or some other sys table perhaps? But sure, I think this is also fine. Can you add a brief comment to each SQL file here, something on the lines of "// Query is operating on literal values. The employee.json file is only used as a valid FROM clause with a multi-row table"

PR #530, opened 2019-01-14
DRILL-3610: Functional tests for TIMESTAMPADD and TIMESTAMPDIFF

abhidotravi:
- Basic functional tests for `TIMESTAMPADD` and `TIMESTAMPDIFF` functions.
- Tests cover different `SQL_TSI_` intervals.

### DIFF
--- a/framework/resources/Functional/timestamp_functions/timestamp_functions.json
+++ b/framework/resources/Functional/timestamp_functions/timestamp_functions.json
@@ -17,5 +17,10 @@
     }
   ],
   "datasources": [
+    {
+      "mode": "cp",
+      "src": "Datasources/Tpch0.01/parquet/orders.parquet",
+      "dest": "Tpch0.01/parquet/orders/orders.parquet"
+    }
   ]
 }


framework/resources/Functional/timestamp_functions/timestamp_functions.json
```diff
@@ -0,0 +1,21 @@
...
+      "schema": "dfs.drilltestdirtpch01parquet",
+      "output-format": "tsv",
+      "expected-file": ".*.e_tsv",
+      "verification-type": [
+        "in-memory"
+      ]
+    }
+  ],
+  "datasources": [
+  ]
```
Agirish: Hey, I understand you are reusing an existing dataset (TPC-H 0.01). But we recommend each test def file to be self sufficient and independent. So can you look up one of the other test suites which uses this dataset / workspace and provide that in the datasources section?

aravi5: I actually referred to `Functional/filter/selectivity/plan` and `Functional/filter/selectivity/negative/plan` to build the definition file and they didn't seem to have anything mentioned for `datasources`.

aravi5: Done.

Agirish: Yeah, I think we might have missed catching it, when adding that suite. But in general, we would want to be able to run any test suite independently. Thanks for fixing this.

PR #534, opened 2019-01-16
Semi Join Test Cases

kedarbcs16:


before-run.sql
```diff
@@ -3,7 +3,7 @@ alter system set `planner.enable_decimal_data_type` = true;
 alter system set `new_view_default_permissions` = '777';
 alter system set `planner.enable_limit0_optimization` = true;
 alter system set `exec.errors.verbose` = true;
-alter system set `planner.memory.max_query_memory_per_node` = 10737418240;
+alter system set `planner.memory.max_query_memory_per_node` = 7737418240;
```
Agirish: Any reason to reduce this? On the regression clusters, we assume 32GB of direct memory. So lets keep this 10GB. In fact might need to increase this to 12 later.

conf/plugin-templates/common/dfs-storage-plugin.template
```diff
@@ -20,6 +20,30 @@
         "writable" : true,
         "defaultInputFormat" : "parquet"
       },
+      "drilltesttpchsf1json": {
```
Agirish: There are already existing workspaces for this data. Can we reuse - or - cleanup?

framework/resources/Functional/semijoin/tpch_sf1Json.json.fail
```diff
@@ -0,0 +1,23 @@
...
+            "schema": "dfs.drilltesttpchsf1json",
+            "output-format": "tsv",
+            "expected-file": ".*.e_tsv",
+            "create-data-in-folder": true,
+	    "verification-type": [
+                "in-memory"
+            ]
+        }
+    ],
+    "datasources": [
```
Agirish: Lets copy over data - not good to leave it blank, as we want these to be independent of other suites

framework/resources/Functional/semijoin/semijoinIn1.sql
```diff
@@ -0,0 +1,3 @@
+alter system set `planner.enable_semijoin` = true;
```
Agirish: Can you include reset after the queries?

PR #536, opened 2019-01-22
DRILL-1248: Add support for using aliases in group by

denysord88:


denysord88:

framework/resources/Datasources/ctas/DRILL-1248/ctas.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+source conf/drillTestConfig.properties
+
+hadoop fs -mkdir -p /drill/testdata/DRILL-1248
```
Agirish: Can we please delete the directory first, to prevent corruption during reruns?

framework/resources/Datasources/ctas/DRILL-1248/ctas.sh
```diff
@@ -0,0 +1,13 @@
...
+
+hadoop fs -put ${DRILL_TEST_DATA_DIR}/Datasources/ctas/DRILL-1248/group_by.tsv /drill/testdata/DRILL-1248
+
+if [ -z "$PASSWORD" ]
+then
+  ${DRILL_HOME}/bin/sqlline -n ${USERNAME} -u "jdbc:drill:schema=dfs.$schema_name;drillbit=${DRILL_STORAGE_PLUGIN_SERVER}" --run=${DRILL_TEST_DATA_DIR}/Datasources/ctas/DRILL-1248/group_by.ddl
```
Agirish: Sorry, can you explain where the $schema_name is set?

framework/resources/Functional/group_by_alias/group_by_index_and_alias.sql
```diff
@@ -0,0 +1 @@
+select intField i, count(intField) ci, strField s, count(strField) cs from dfs.drillTestDir.`DRILL-1248/group_by_parquet` group by 1, s;
```
Agirish: If we group by 'i' & 1 (duplicate), will Drill catch during parsing?

framework/resources/Datasources/ctas/DRILL-1248/ctas.sh
denysord88: $schema_name is in the framework/resources/Functional/group_by_alias/group_by_alias.json
"schema": "dfs.drillTestDir"

framework/resources/Datasources/ctas/DRILL-1248/ctas.sh
denysord88: I will not delete a source directories. This folder may be used by other test suites in the future, and they may be corrupted by multithreading.

Agirish: So for that, please use the 'cp' command in the test framework - it'll take care of copying over source data. The gen script should be re-runnable without failures. Here, it will fail in the below copy step.

framework/resources/Datasources/ctas/DRILL-1248/ctas.sh
Agirish: Sorry still not clear - I mean where is it exported? I understand that the "schema_name" here has same value as "schema" in test def file.. but how does this script get that reference? Sorry could be a silly question - I guess I'm missing something

framework/resources/Functional/group_by_alias/group_by_index_and_alias.sql
Agirish: Thanks for clarifying!

Agirish:
Thanks for addressing review comments. I'll take another look and have it merged if it's all good. I'm assuming you've tested this PR and there were no regressions / issues.

framework/resources/Datasources/ctas/DRILL-1248/group_by.ddl
```diff
@@ -0,0 +1,4 @@
+DROP TABLE IF EXISTS dfs.drillTestDir.`DRILL-1248/group_by_parquet`;
+create table dfs.drillTestDir.`DRILL-1248/group_by_parquet`
```
Agirish: A minor request for change - can we have the SQL statements in one case format? It's all caps above for DROP TABLE and all lower case for CREATE TABLE

framework/resources/Datasources/ctas/DRILL-1248/ctas.sh
denysord88: Done.

PR #542, opened 2019-02-11
Update framework to ignore most formatting differences between

rhou1:
Apache JDBC and Simba JDBC results.
Added ability to include Jira number to failed tests.

### DIFF
--- a/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
@@ -19,6 +19,7 @@ package org.apache.drill.test.framework;
 
 import java.math.BigDecimal;
 import java.sql.Types;
+import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -101,14 +128,17 @@ public String toString() {
       int type = (Integer) (types.get(i));
       if (Simba && (type == Types.VARCHAR)) {
```
kkhatua: This should ideally be set as `isSimba`

framework/src/main/java/org/apache/drill/test/framework/ColumnList.java
```diff
@@ -101,14 +128,17 @@ public String toString() {
...
+        // If the field has a JSON string or list, then remove newlines
+        // so that each record fits on a single line in the actual output
+        // file.  json/json_kvgenflatten/kvgen/kvgen7.q
+        // Also, sometimes Simba returns varchars with white spaces
+        // in different places.  Remove the white space.
+        // convert/convert32.q
         if ((s1.length() > 0) &&
            ((s1.charAt(0) == '{') || (s1.charAt(0) == '[')) ) {
           s1 = Utils.removeNewLines(s1);
+          s1 = removeStringSpace(s1);
```
kkhatua: There might be libraries that do the reverse of _prettifying_ JSON string, including stripping down extra spaces. Consider having a single Util method that does something like `s1 = Utils.compactJson(s1);`

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -398,7 +403,24 @@ private static String getExpectedFile(String queryFile, String queryFileExt,
...
+                          file.getName().endsWith(".sjdbc.fail")) {
+                        FOUNDFAILED = true;
+                        break;
+                      }
+                    }
+                  }
+                  if (FOUNDFAILED) {
+                    return null;
+                  } else {
+                    return Filename;
```
kkhatua: Variables need to be in lower camel case

PR #550, opened 2019-03-07
DRILL-6962: Added tests to verify that coalesce function doesn't throw exception if use it with non-existing columns.

agozhiy:


framework/resources/Functional/functions/functions.json
```diff
@@ -0,0 +1,22 @@
+{
+  "testId": "functions",
```
Agirish: Hey @agozhiy I would suggest retaining the new "functions" dir and creating a sub dir for coalesce function. The tests here can be moved inside that. And correspondingly the testID here can be updated to be more unique to it.

agozhiy: Done.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -554,6 +554,9 @@ public static String getSqlResult(ResultSet resultSet) throws SQLException {
               values.add(null);
             }
             break;
+          case Types.NULL:
+            values.add(null);
+            break;
```
Agirish: Hey @agozhiy , can you please have any framework code changes in a separate PR? I think it will help with maintenance.

agozhiy:
Done. Opened a PR:
https://github.com/mapr/drill-test-framework/pull/560

PR #554, opened 2019-03-12
MD-5043 S3 Test Automation

denysord88:
Added Minio server to test Drill with Amazon S3 API.
Added tests to the Drill with S3 storage.

### DIFF
--- a/conf/plugin-templates/common/s3-storage-plugin.template
+++ b/conf/plugin-templates/common/s3-storage-plugin.template
@@ -1,0 +1,74 @@
+{
+  "name" : "s3",
+  "config" : {
+    "type": "file",
+    "enabled": true,
+    "connection": "s3a://tstbucket",
+    "config": {
+      "fs.s3a.endpoint": "http://localhost:9000",
+      "fs.s3a.connection.ssl.enabled": "false",
+      "fs.s3a.path.style.access": "true",
+      "fs.s3a.access.key": "ZYVR33VM12IAHETIJYF9",
+      "fs.s3a.secret.key": "BjGb9WpNd74t5T7FcBlXRgUcSyq505G598T8Yxg6",
+      "fs.s3a.impl.disable.cache": "false",
+      "fs.s3a.connection.maximum": "30"
+    },
+    "workspaces": {
+      "tmp": {
+        "location": "/tmp",
+        "writable": true,
+        "defaultInputFormat": null,
+        "allowAccessOutsideWorkspace": false
+      },
+      "root": {
+        "location": "/",
+        "writable": false,
+        "defaultInputFormat": null,
+        "allowAccessOutsideWorkspace": false
+      }
+    },
+    "formats": {
+      "csv": {
+        "type": "text",
+        "extensions": [
+          "csv"
+        ],
+        "delimiter": ","
+      },
+      "tsv": {
+        "type": "text",
+        "extensions": [
+          "tsv"
+        ],
+        "delimiter": "\t"
+      },
+      "parquet": {
+        "type": "parquet"
+      },
+      "json": {
+        "type": "json",
+        "extensions": [
+          "json"
+        ]
+      },
+      "avro": {
+        "type": "avro"
+      },
+      "csvh": {
+        "type": "text",
+        "extensions": [
+          "csvh"
+        ],
+        "extractHeader": true,
+        "delimiter": ","
+      },
+      "image": {
+        "type": "image",
+        "extensions": [
+          "jpg",
+          "png"
+        ]
+      }
+    }
+  }
+}
--- a/framework/resources/Datasources/s3/minio/stop_mn.sh
+++ b/framework/resources/Datasources/s3/minio/stop_mn.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+if [ ! -z "$(top -bn1 |grep minio |grep -oE '^[ ]*[0-9]+' |grep -oE '[0-9]+')" ]
+then
+  kill -9 $(top -bn1 |grep minio |grep -oE '^[ ]*[0-9]+' |grep -oE '[0-9]+')
+fi
--- a/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
@@ -716,6 +716,11 @@ public class TestDriver {
 		}
 	} 
 	connectionPool.releaseConnection(connection);
+
+    // Stop Apache Minio server if it was started for the S3 storage
+    if(cmdParam.excludeDependencies == null || !cmdParam.excludeDependencies.contains("s3")) {
+      Utils.stopMinio();
+    }
   }
 
   private void prepareData(List<DrillTestCase> tests) throws Exception {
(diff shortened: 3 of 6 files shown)


denysord88:

conf/plugin-templates/common/s3-storage-plugin.template
```diff
@@ -0,0 +1,124 @@
...
+  "config" : {
+    "type": "file",
+    "enabled": true,
+    "connection": "s3a://tstbucket",
+    "config": {
+      "fs.s3a.endpoint": "http://localhost:9000",
+      "fs.s3a.connection.ssl.enabled": "false",
+      "fs.s3a.path.style.access": "true",
+      "fs.s3a.access.key": "ZYVR33VM12IAHETIJYF9",
+      "fs.s3a.secret.key": "BjGb9WpNd74t5T7FcBlXRgUcSyq505G598T8Yxg6",
```
Agirish: Are these okay to be public?

denysord88: It isn't a real Amazon S3 credentials. I just choose this random keys and hardcode it for the Minio server.

conf/plugin-templates/common/s3-storage-plugin.template
```diff
@@ -0,0 +1,124 @@
...
+      },
+      "csvh": {
+        "type": "text",
+        "extensions": [
+          "csvh"
+        ],
+        "extractHeader": true,
+        "delimiter": ","
+      },
+      "image": {
```
Agirish: If we are not using all format plugins, I'd suggest removing everything except the formats you are and will be interested in.

denysord88: Removed 'psv' and 'sequencefile' formats.

Agirish: Thanks!

framework/resources/Datasources/s3/fragments/fragments.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+source conf/drillTestConfig.properties
+
+mkdir ${DRILL_TEST_DATA_DIR}/Datasources/s3/minio/minio_data/tstbucket/tmp/big_files
+
+tar -xzmvf ${DRILL_TEST_DATA_DIR}/Datasources/ctas/decimal/fragments/decimal_big.tar.gz -C ${DRILL_TEST_DATA_DIR}/Datasources/s3/minio/minio_data/tstbucket/tmp/big_files
```
Agirish: Sorry, can't seem to figure out - where are these tar.gz files coming from?

framework/resources/Functional/s3/fpushdown/fpushdown.json
```diff
@@ -0,0 +1,23 @@
...
+      "query-file": ".*.sql.explain",
+      "schema": "s3.tmp",
+      "output-format": "tsv",
+      "expected-file": ".*.plan",
+      "verification-type": [
+        "regex"
+      ]
+    }
+  ],
+  "datasources": []
```
Agirish: Don't these need any data to work on? If it's depending on some other test suite, then please add those copy / gen scripts here as well. We want to make sure each test suite can be executed independently.

framework/src/main/java/org/apache/drill/test/framework/TestDriver.java
```diff
@@ -716,6 +716,11 @@ private void teardown() {
 		}
 	} 
 	connectionPool.releaseConnection(connection);
+
+    // Stop Apache Minio server if it was started for the S3 storage
+    if(cmdParam.excludeDependencies == null || !cmdParam.excludeDependencies.contains("s3")) {
```
Agirish: Hey my suggestion is to get these framework changes in separately in a different PR. That way we can keep tests and src independent.

denysord88: Ok. I'm closing this PR and will create 2 separate PR's for tests and logic.

Agirish:
Actually we can reopen this PR. Just remove changes for code here. And a new PR for the code changes?

denysord88:
Done. I removed S3 tests from this PR.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -667,6 +694,34 @@ private static boolean updateDrillStoragePlugin(String filename, String pluginTy
...
+   *
+   * @param filename
+   *          name of file containing drill storage plugin
+   * @param pluginType
+   *          type of plugin; e.g.: "dfs", "cp"
+   * @param enabled
+   *          value of the "enabled" field in a storage plugin
+   * @return true if operation is successful
+   */
+  private static boolean changeDrillStoragePluginStatus(String filename, String pluginType, String fsMode, boolean enabled) {
```
Agirish: Hey @denysord88 , Is this used only for enabling / disabling storage plugins? 

If so, i'm trying to understand the steps below which replace maprfs and location. I'm not sure why that's needed and I don't think this is generic enough for any storage plugin. If one intends to use hdfs for example, or if the plugin doesn't have the location field. 

Can we have this method exclusively enable / disable plugins?

framework/resources/Datasources/s3/minio/stop_mn.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+if [ ! -z "$(top -bn1 |grep minio |grep -oE '^[ ]*[0-9]+' |grep -oE '[0-9]+')" ]
+then
+  kill -9 $(top -bn1 |grep minio |grep -oE '^[ ]*[0-9]+' |grep -oE '[0-9]+')
```
Agirish: Is there no clean way to stop the server?

denysord88: Minio server is running in the background by running command "nohup ... &".
There is no way to send Minio any shutdown commands.
If we don't run Minio server in the background, the only way to stop it is to close bash window or send Ctrl+C key combination to this bash window.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
denysord88: We need this method to disable s3 storage plugin after all tests are done.
Since we are stopping Minio server at the end of the test run, we can't leave S3 storage plugin enabled. It will cause SQL Exception before the next tests will be started.

I didn't want to overload current **_updateDrillStoragePlugin(String filename, String pluginType, String fsMode)_** method because disabling storage plugin will be used only for the s3 storage plugin at the end of the test run if S3 tests aren't excluded.

framework/resources/Datasources/s3/minio/stop_mn.sh
Agirish: Sure, thanks for clarifying.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
Agirish: Still not clear why we are doing the other modifications if we are only interested in disabling the plugins. And I think we can then have a simpler disablePlugin method which can be used to disable any plugin.

denysord88: Done. Updated this method to the disableStoragePlugin(String filename, String pluginType) and removed all replacements instead of the "enabled" option.

PR #555, opened 2019-03-12
Enable stats tests

anishareddyv:
Done testing also.
------------------------------------------------------------------------
ITERATION SUMMARY
------------------------------------------------------------------------
Total Tests                  : 16
Passing Tests                : 16
Failing Tests                : 0

> Execution Failures         : 0
> Data Verification Failures : 0
> Plan Verification Failures : 0
> Timeout Failures           : 0

Iteration 1 (of 1) duration: 22.70 s
------------------------------------------------------------------------

------------------------------------------------------------------------
TEARDOWN
------------------------------------------------------------------------
> Executing queries


framework/resources/Functional/table_stats/stats/positive/meta_stats.sql
```diff
@@ -1 +1,2 @@
-explain plan including all attributes for select l_orderkey,l_partkey,l_shipdate,l_shipmode from `table_stats/lineitem_parquet_partitioned_hive_date` where dir0 = 'dt=2014-10-22' and l_shipmode='AIR';
+set `planner.statistics.use` = true;
```
kedarbcs16: if there are even no of queries you have to add a --@test to the query u want the result on.
My suggestion would be to reset `planner.statistics.use` when u set them

Agirish: Yes, I would recommend first doing a SET and then a RESET. Here there is actually just one query to be tested.

anishareddyv: Did a reset.

Agirish:
Please make sure the tests pass on apache/drill master before merging :)

PR #558, opened 2019-03-15
Statistics: Tests to verify simple query plans

anishareddyv:
This PR contains: 

1. tests for plan verification of simple queries.
2. Tests for stats on nested scalar cols


framework/resources/Functional/table_stats/stats/query/query.json
```diff
@@ -0,0 +1,31 @@
+{
+  "testId": "Table_Stats_Query",
```
Agirish: Test ID has to be unique. This is because all tests are collected and are identified as <testID>.<queryFileName>. I checked that your Test IDs here matches the one in the dir `positive`

anishareddyv: Oh ok I did not know that. I changed the test ID of positive json

Agirish: Thanks for updating!

anishareddyv:
Yes it did run successfully.

PR #564, opened 2019-03-25
DRILL-6976: Added tests to verify that an empty function result in a …

agozhiy:
…subquery no longer throws an exception.


framework/resources/Functional/functions/split/split.json
```diff
@@ -0,0 +1,22 @@
+{
+  "testId": "coalesce",
```
Agirish: I think coalesce was used? Change to "split"?

agozhiy: Fixed.

PR #573, opened 2019-04-08
Added command line parameter "--jvm" to pass Java options to the test…

agozhiy:
… runs

It can be used to enable remote debug, for example:
./bin/run_tests -s Functional -g functional -d --jvm "-Xdebug -Xnoagent -Xrunjdwp:transport=dt_socket,address=50000,server=y,suspend=y"

### DIFF
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -4,6 +4,29 @@ bin=`cd "$bin">/dev/null; pwd`
 source conf/drillTestConfig.properties
 cd ${bin}/../
 
+default_debug_string="-Xdebug -Xnoagent -Xrunjdwp:transport=dt_socket,address=50000,server=y,suspend=y"
+
+ARGS=()
+while [[ $# > 0 ]]
+do
+  arg="$1"
+  case "$arg" in
+  --jvm)
+    shift
+    DRILL_TEST_FRAMEWORK_JAVA_OPTS="$DRILL_TEST_FRAMEWORK_JAVA_OPTS $1"
+    shift
+    ;;
+  --debug)
+    DRILL_TEST_FRAMEWORK_JAVA_OPTS="$DRILL_TEST_FRAMEWORK_JAVA_OPTS $default_debug_string"
+    shift
+    ;;
+  *)
+    ARGS+=("$1")
+    shift
+    ;;
+  esac
+done
+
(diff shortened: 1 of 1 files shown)


Agirish:
Since we only have a few main possibilities - debug being one of them, why not provide a flag called --debug or similar which starts test framework in debug mode? We can keep the default values for port address and such. Anyone can tweak them if needed. This way noone needs to actually remember the long debug string.

agozhiy:
I wanted to implement is the same way it's implemented in Drill. Please see Paul's comments here: [DRILL-6270](https://issues.apache.org/jira/browse/DRILL-6270?focusedCommentId=16439657&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16439657), https://github.com/apache/drill/pull/1210

Also, you don't need to remember debug string, you can use aliases or functions, for example:
``` bash
run_with_debug () {
    local cmd=$1
    eval '$cmd --jvm "-Xdebug -Xnoagent -Xrunjdwp:transport=dt_socket,address=50000,server=y,suspend=y"'
}
```
That way you avoid passing additional parameters like port or suspend.

Agirish:
But you do have to remember and pass the whole string here.. am I missing something? My suggestion was to have meaningful defaults and start in debug mode by a simple flag like --debug. Anyone and open the file and override if they want to change port or something. Where is the debug function you shared above?

agozhiy:
My point is that you can store such function in your .bashrc file, and then run tests like this:
``` bash
run_with_debug bin/run_tests -s Functional -g functional -d
```
Yes, anyone can open the file and actually add the whole debug string, but that is not very convenient. 
Also, general --jvm parameter can be used not only for debugging but any other java options (heap size, timezone etc)
As a compromise I suggest to add both --jvm and --debug with the default string. What do you think?

Agirish:
I see the point for it to be generic, although I don't completely agree with it. So, yes.. please add both options if possible. I'd appreciate it. I think for most cases the --debug flag will be easy to use and sufficient. For advanced / custom scenarios one can use your --jvm parameter.

agozhiy:
Done.

agozhiy:
I tested it manually and tests were passed.

PR #574, opened 2019-04-08
Added the Framework version check before the test runs

agozhiy:
It is useful when analyzing Jenkins test job results

### DIFF
--- a/framework/pom.xml
+++ b/framework/pom.xml
@@ -204,6 +204,31 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>pl.project13.maven</groupId>
+        <artifactId>git-commit-id-plugin</artifactId>
+        <version>2.2.5</version>
+        <executions>
+          <execution>
+            <id>get-the-git-infos</id>
+            <goals>
+              <goal>revision</goal>
+            </goals>
+          </execution>
+        </executions>
+
+        <configuration>
+          <dotGitDirectory>${project.basedir}/../.git</dotGitDirectory>
+          <generateGitPropertiesFile>true</generateGitPropertiesFile>
+          <generateGitPropertiesFilename>${project.basedir}/../git.properties</generateGitPropertiesFilename>
+          <failOnNoGitDirectory>false</failOnNoGitDirectory>
+          <failOnUnableToExtractRepoInfo>false</failOnUnableToExtractRepoInfo>
+          <gitDescribe>
+            <always>false</always>
+            <forceLongFormat>true</forceLongFormat>
+          </gitDescribe>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
   <profiles>
--- a/framework/src/main/java/org/apache/drill/test/framework/Utils.java
+++ b/framework/src/main/java/org/apache/drill/test/framework/Utils.java
@@ -53,6 +53,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.io.Resources;
 
 import org.apache.http.HttpResponse;
 import org.apache.http.NameValuePair;
@@ -1069,5 +1070,26 @@ public class Utils {
         return false;
       }
     return true;
+  }
+
+  public static String getFrameworkVersion() {
+    String commitID = "";
+    String commitAuthor = "";
+    String commitEmail = "";
+    String commitMessage = "";
+    try {
+      URL u = Resources.getResource("git.properties");
+      if (u != null) {
+        Properties p = new Properties();
+        p.load(Resources.asByteSource(u).openStream());
+        commitID = p.getProperty("git.commit.id");
+        commitAuthor = p.getProperty("git.commit.user.name");
+        commitEmail = p.getProperty("git.commit.user.email");
+        commitMessage = p.getProperty("git.commit.message.short");
+      }
+    } catch (IOException | IllegalArgumentException e) {
+      LOG.warn("Failure while trying to load \"git.properties\" file.", e);
+    }
+    return String.format("Commit: %s\nAuthor: %s <%s>\n\n%s", commitID, commitAuthor, commitEmail, commitMessage);
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ conf/drillTestConfig.properties
 *.class
 **/target/*
 logs/*
+git.properties
 framework/resources/Datasources/s3minio/minio/minio_data/.minio.sys
 framework/resources/Datasources/s3minio/minio/minio_data/tstbucket/tmp/ppruning/
 framework/resources/Datasources/s3minio/minio/minio_data/tstbucket/tmp/gitignore
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -34,7 +34,7 @@ done
 # use the following line when testing apache drill JDBC driver
 if [[ $HADOOP_VERSION == *"mapr"* ]]
 then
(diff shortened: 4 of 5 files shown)


framework/scripts/framework_version.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+git log -n1 --pretty=format:'Commit: %H%nAuthor: %aN <%ae>%n%n%B%n'
```
Agirish: I don't think this would work if someone downloads the test framework as a .tar.gz file from the releases section of GitHub. And this may fail then. 

I'm just thinking out possibilities here.. If we can have a git.properties file similar to Drill, it may help?

aravi5: [Git commit id plugin in Drill](https://github.com/apache/drill/blob/master/pom.xml#L539) is what @Agirish is talking about. We could try it out.

agozhiy: Need some time to try it out.

agozhiy: @Agirish, @aravi5, 
As it appears, in cases when .git folder is absent git commit plugin also won't work. Although I can do this, and it appears to me more flexible than running a command from console.

Agirish: @agozhiy, I see. But we expect the git commit plugin to be only run when building the framework - which is expected to be via git clone.

So there are two ways people can use the drill-test-framework:
(1) git clone - build and then run tests. Here the git plugin approach should work.
(2) use a tar ball of the framework. Here, the expectation is that the tarball already contains the git.properties file. When building the tar ball, it's expected to be via step (1). 

On GitHub, I've been doing periodic releases (coinciding with Drill versions). But I think that just includes the src - not a built version. We should find a space to build and host the tar ball releases.

Agirish: Also, the git commit plugin approach appears to be more standard, so I think it would be good to go that route than running a command.

Agirish: Update: Looks like we can host the built tar-ball on GitHub under releases, along with the source. They need to be uploaded. Starting next release, we can do that.

agozhiy: I agree.
And as I see, we need Drill jdbc driver to build framework, wouldn't it be a problem for creating tar-balls?

Agirish: Actually for that, I'm thinking by default we should just pick up the latest release version of Apache Drill in the pom dependency - rather than using the system path we have today. 

But we should have an option to override it in our Jenkins test env, as we also want to test the latest JDBC Driver. This is mainly to make sure any changes in JDBC layer are also caught. Plus we historically had issues with compatibility across versions of Drill - which is a lot better today.

This way anyone can try out the test framework without having a runtime dependency. And for most tests that should be sufficient.

agozhiy:
Rewrote using git commit id maven plugin.
@Agirish, could you please review?

framework/pom.xml
```diff
@@ -204,6 +204,31 @@
...
+            <goals>
+              <goal>revision</goal>
+            </goals>
+          </execution>
+        </executions>
+
+        <configuration>
+          <dotGitDirectory>${project.basedir}/../.git</dotGitDirectory>
+          <generateGitPropertiesFile>true</generateGitPropertiesFile>
+          <generateGitPropertiesFilename>${project.basedir}/../conf/framework.git.properties</generateGitPropertiesFilename>
```
Agirish: I think we can just call this as `git.properties` and keep it in the root level of the repository. That's where it's usually present.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -1068,4 +1069,25 @@ public static boolean matches(String actual, String expected) {
       }
     return true;
   }
+
+  public static String getFrameworkVersion() {
+    String commit = "";
+    String commitAuthor = "";
+    String commitMail = "";
```
Agirish: Call this commitEmail?

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -1068,4 +1069,25 @@ public static boolean matches(String actual, String expected) {
       }
     return true;
   }
+
+  public static String getFrameworkVersion() {
+    String commit = "";
```
Agirish: Call this commitID?

Agirish:
Thanks @agozhiy . Added some minor commits. Other wise looks good. Can you share a sample output? So we can suggest any formatting fixes

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -1068,4 +1069,25 @@ public static boolean matches(String actual, String expected) {
...
+      if (u != null) {
+        Properties p = new Properties();
+        p.load(Resources.asByteSource(u).openStream());
+        commit = p.getProperty("git.commit.id");
+        commitAuthor = p.getProperty("git.commit.user.name");
+        commitMail = p.getProperty("git.commit.user.email");
+        commitMessage = p.getProperty("git.commit.message.short");
+      }
+    } catch (IOException | IllegalArgumentException e) {
+      LOG.warn("Failure while trying to load \"framework.git.properties\" file.", e);
```
aravi5: In case of Exception or `framework.git.properties` not available, the **DRILL TEST FRAMEWORK VERSION** would look something like
```
Commit: null
Author: null <null>


null
```

framework/pom.xml
agozhiy: There is a problem here. It is conflicting with git.properties from the Drill JDBC driver. And for some reason it was not working when I added it to the root. Maybe because the classpath we set in run_tests: 
`-cp conf/:${DRILL_HOME}/jars/jdbc-driver/drill-jdbc-all-${DRILL_VERSION}.jar:framework/target/framework-1.0.0-SNAPSHOT-shaded.jar`

Of course I can open the properties not as a resource but as a regular file, but I don't like the idea...

agozhiy:
Here is the output:
```
+ ./bin/run_tests -s Functional,External/mapr-drill-tests/Testcases -g functional -t 1200 -d -n 40 -x hbase,mysql,mysql,hbase
------------------------------------------------------------------------
DRILL TEST FRAMEWORK VERSION
------------------------------------------------------------------------
Commit: eb564b142f3574d9668ddc9896f3d6c155f6be85
Author: Anton Gozhiy <anton5813@gmail.com>

Added the Framework version check before the test runs
------------------------------------------------------------------------

------------------------------------------------------------------------
JDBC DRIVER METADATA CATALOG
------------------------------------------------------------------------
```

framework/pom.xml
Agirish: Do you think swapping the cp order in the run_tests will work? I think we should try and understand what's causing the issue. Using resource is good. Confirming to standard names would also be good.

agozhiy: But we would have 2 different git.properties in one classpath, and which one is picked is only a question of order... Is it ok?

Agirish: In Apache projects such as Drill which also have other dependent Apache projects in it - each of which may have git.properties, how do you think it's handled?

agozhiy: Well, I'll try to figure it out.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
agozhiy: Done.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
agozhiy: Done.

framework/pom.xml
agozhiy: Added git.properties to the project root, added the project root to the classpath in run_tests.

agozhiy:
@Agirish, I addressed your comments, please review.

framework/src/main/java/org/apache/drill/test/framework/Utils.java
```diff
@@ -1070,4 +1071,25 @@ public static boolean matches(String actual, String expected) {
...
+      if (u != null) {
+        Properties p = new Properties();
+        p.load(Resources.asByteSource(u).openStream());
+        commitID = p.getProperty("git.commit.id");
+        commitAuthor = p.getProperty("git.commit.user.name");
+        commitEmail = p.getProperty("git.commit.user.email");
+        commitMessage = p.getProperty("git.commit.message.short");
+      }
+    } catch (IOException | IllegalArgumentException e) {
+      LOG.warn("Failure while trying to load \"framework.git.properties\" file.", e);
```
Agirish: This needs to be updated?

agozhiy: Yes, missed it. Fixed.

PR #575, opened 2019-04-10
DRILL-7063: Update addremove_files.sh

agirish:
Update the metadata cache file names with .v4 extension.


Agirish:
Tried this fix on CV test cluster - fixes the data gen failures which were seen. However, it looks like the tests will also need to be updated to read from the split metadata cache files (previously it was assuming there is a single cache file). Some of the tests which I saw are disabled, so Im guessing it should be no effort for now. But when testing the new metadata cache files, these tests need to be updated and enabled.

PR #577, opened 2019-04-11
Updated the schema name in generation scripts

denysord88:


denysord88:
@Agirish Could you please review the PR?

PR #584, opened 2019-05-13
DRILL-7050 - Testing subqueries in the SELECT statement.

denysord88:


denysord88:

Agirish:
@denysord88 , can you please share why we are using both cp and dfs storage plugins? Also the dfs table has same data as the one in cp - so trying to understand.

Otherwise, looks good. +1

denysord88:
> @denysord88 , can you please share why we are using both cp and dfs storage plugins? Also the dfs table has same data as the one in cp - so trying to understand.
> 
> Otherwise, looks good. +1

I just want to check JOIN tables from different workspaces in the subquery. I think it may cause some issues and it will be good to have this test in our framework.

Agirish:
I'd like to avoid unnecessary data
Gen scripts if possible. So if this can be reworked to use an existing DFS table, it would be good. It's just easy for overall maintenance. And for overall storage footprint.

Agirish:
On second thoughts, since this is a small dataset, let's get it in. But please be mindful Ruth future PRs. More scripts and more data is more maintenance

PR #585, opened 2019-05-16
DRILL-6974: Three new tests for 'set' command 

Atri0:
Checks 
1) Are "ALTER SYSTEM / SESSION " modifier for SET command still works and works correctly
2) SET 'PROPERTY'; - should show current value of property
3) Is it still possible to update boolean / string / numeric values for properties

### DIFF
--- a/framework/resources/Functional/query_parser/DRILL-6974_alter_session_set_command_works.sql
+++ b/framework/resources/Functional/query_parser/DRILL-6974_alter_session_set_command_works.sql
@@ -1,0 +1,3 @@
+set  store.kafka.poll.timeout = 201;
+alter session set store.kafka.poll.timeout;
+reset store.kafka.poll.timeout;
--- a/framework/resources/Functional/query_parser/DRILL-6974_set_returns_current_value_of_property.sql
+++ b/framework/resources/Functional/query_parser/DRILL-6974_set_returns_current_value_of_property.sql
@@ -1,0 +1,3 @@
+set `store.format`='json';
+alter system set `store.format`;
+reset `store.format`;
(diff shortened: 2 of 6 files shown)


framework/resources/Functional/query_parser/DRILL-6974_set_returns_current_value_of_property.sql
```diff
@@ -0,0 +1,3 @@
+set `store.format`='json';
+alter system set `store.format`;
```
Agirish: This could be risky. System option changes can impact other tests. Here CTAS tests may be forced to write to JSON, depending on when this test is executed.

framework/resources/Functional/query_parser/DRILL-6974_alter_session_set_command_works.sql
```diff
@@ -0,0 +1,3 @@
+set  store.kafka.poll.timeout = 201;
+alter session set store.kafka.poll.timeout;
```
Agirish: Without value, is the set command valid? I haven't looked at recent changes

framework/resources/Functional/query_parser/DRILL-6974_set_returns_current_value_of_property.sql
Atri0: At current test we do not change system properties. We change session properties and verify that system property does not change. Also, at the third line we return session property to its default value.

framework/resources/Functional/query_parser/DRILL-6974_alter_session_set_command_works.sql
Atri0: SET "PROPERTY_NAME" ;   (without value) is new functionality that was implemented in scope of DRILL-6974. It returns current value of property that is set at current time.

Agirish: Thanks for the clarification!

PR #594, opened 2019-08-05
Added an option for float / double precision that would help to solve…

agozhiy:
… precision related test failures


agozhiy:
@Agirish, could you please review? 
This PR does the same as Kedar's except that the options are added to test suite configs json instead of command line parameters. It doesn't affect existing tests as the options are optional and have default values. A new type of exception can be added in a separate PR if needed.

Agirish:
@agozhiy , taking a look...

Agirish:
Thanks @agozhiy !!

PR #600, opened 2019-10-28
DRILL-7418: Fixed baselines due to optimizations

agozhiy:
After DRILL-7418, we don't enumerate files used during direct count optimization, only selection root and number of files are listed.
For Hive selection root is unavailable thus only number of files will be present in plan.

Need to be merged after the following PR:
https://github.com/apache/drill/pull/1883


arina-ielchiieva:
Please merge the PR, DRILL-7418 was merged into Apache master.

PR #603, opened 2020-04-03
DRILL-7615: Added tests to verify that decimal columns work correctly…

dkondriu:
DRILL-7615: Added tests to verify that decimal columns work correctly with UNION operator

Also renamed existing tests according to the project naming convention


agozhiy:
+1, LGTM.

